PR #137, opened 2019-04-09
(feat): add mac support

thomaszurkan-optimizely:


severdia:
OK, the password works now. The site was down for a bit while they added/fixed SSL, but fixed now.

PR #202, opened 2019-05-23
(chore): fix readme doc links

jaeopt:


jaeopt:
> Can't reach url.

Recovered the old doc links and github links before doc and cocoapods released
Also fixed a couple of format errors too

PR #210, opened 2019-05-29
(feat): SwiftLint added.

yasirfolio3:
There are 3491 issue found with SwiftLint.  These should all be cleaned up in this PR.


yasirfolio3:
> looks good!
> Can we add .swiftlint.yml file so we can customize rules. For now, we have too many errors/warnings. I guess some of them we can off for now.

.swiftlint.yml file added with most widely used rules. Fixes also made to the sdk according to the linter recommendations.

PR #218, opened 2019-06-04
add tests for Data Model coverage

jaeopt:


OptimizelySDK/Data Model/Audience/AttributeValue.swift
```diff
@@ -81,10 +81,13 @@ enum AttributeValue: Codable, Equatable, CustomStringConvertible {
             return
         }
         
-        if let value = try? container.decode(Int64.self) {
+        // Int values are all filtered as Double, so this Int parsing is not needed here
```
thomaszurkan-optimizely: remove the comments and commented out code.

OptimizelySDK/Data Model/Audience/ConditionHolder.swift
```diff
@@ -95,9 +95,6 @@ extension Array where Element == ConditionHolder {
     }
     
     func evaluate(op: LogicalOp, project: ProjectProtocol?, attributes: OptimizelyAttributes?) throws -> Bool {
-        guard self.count > 0 else {
```
thomaszurkan-optimizely: why is this removed?

OptimizelySDK/Data Model/Audience/UserAttribute.swift
```diff
@@ -97,7 +97,7 @@ extension UserAttribute {
 
         // invalid match - parsed for forward compatibility only (but evaluation fails)
         guard let matchFinal = matchSupported else {
-            throw OptimizelyError.userAttributeInvalidMatch(self.match ?? "empty")
+            throw OptimizelyError.userAttributeInvalidMatch(self.match!)
```
thomaszurkan-optimizely: there is no guarantee that match is not nil.  so either bind it or default it but not bang.

OptimizelySDK/Data Model/DispatchEvents/EventForDispatch.swift
```diff
@@ -35,7 +35,13 @@ import Foundation
 }
 
 extension EventForDispatch {
+    
     override public var description: String {
         return "[url] \(url) (" + (String(data: body, encoding: .utf8) ?? "UNKNOWN") + ")"
     }
+
```
thomaszurkan-optimizely: why did you move this?

OptimizelySDK/OptimizelyTests/OptimizelyTests-DataModel/ConditionLeaf.swift
```diff
@@ -0,0 +1,23 @@
+/****************************************************************************
```
thomaszurkan-optimizely: why is this class here?

OptimizelySDK/Data Model/Audience/ConditionHolder.swift
jaeopt: This condition is covered by the caller, so the guard-else condition never happens

OptimizelySDK/Data Model/Audience/UserAttribute.swift
jaeopt: nil match case is filtered out by "matchSupported", so safe for bang to remove unused code.
Will refactor the code to make safety more clear.

OptimizelySDK/Data Model/DispatchEvents/EventForDispatch.swift
jaeopt: Looks like redundant to "isEqual" above. WIl be removed.

OptimizelySDK/OptimizelyTests/OptimizelyTests-DataModel/ConditionLeaf.swift
jaeopt: Good catch. Looks like an error introduced while merging. Will clean up xcode project.

OptimizelySDK/OptimizelyTests/OptimizelyTests-DataModel/ConditionHolderTests_Evaluate.swift
```diff
@@ -162,3 +177,39 @@ extension ConditionHolderTests_Evaluate {
...
+        let model: ConditionHolder = try! OTUtils.model(from: [["and", userAttributeData]])
+        do {
+            _ = try model.evaluate(project: project, attributes: attributeData)
+            XCTAssert(false, "only operator or leaf node can be the first item")
+        } catch {
+            XCTAssert(true)
+        }
+    }
+    
+    func testEvaluate_OperatorOnly() {
```
thomaszurkan-optimizely: Where is the error thrown here?  This appears to now hit the  self[1...] where it would throw an exception

jaeopt: I guess you expect out-of-bound exception here. It does not happen since the array size is 1 here ( "and" is not filtered out yet).

PR #225, opened 2019-06-11
add tests for Optimizely coverage

jaeopt:


OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_Invalid.swift
```diff
@@ -116,3 +127,19 @@ class OptimizelyClientTests_Invalid: XCTestCase {
     }
 
 }
+
+// MARK: - Invalid Datafile
+
+extension OptimizelyClientTests_Invalid {
+    
+//    func testActivate_WhenManagerNonInitialized() {
+//        do {
```
thomaszurkan-optimizely: what's this?

OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_ObjcOthers.m
```diff
@@ -0,0 +1,272 @@
...
+- (void)flushEvents {
+    return;
+}
+@end
+
+@implementation OptimizelyClientTests_ObjcOthers
+
+// MARK: - Others
+
+- (void)testCustomNotificationCenter_Activate {
```
thomaszurkan-optimizely: this seems incorrectly named CustomNotificationCenter_Activate?

OptimizelySDK/Customization/DefaultEventDispatcher.swift
```diff
@@ -75,9 +75,9 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
     open func dispatchEvent(event: EventForDispatch, completionHandler: DispatchCompletionHandler?) {
         dataStore.save(item: event)
         
-        // TODO: use or clean up completionHandler
-
         setTimer()
+        
+        completionHandler?(.success(event.body))
```
thomaszurkan-optimizely: this doesn't seem like the correct approach.  maybe something  better is to store a list of all callbacks without a retain and then call them on anytime an event is actually sent.

jaeopt: The current way of calling back on queueing makes sense to me:
- We do not have any case for queueing failure for now, but this call can be used for notification of "event queued successfully and it'll be delivered ok"
- Event delivery can be delayed arbitrary time. Keeping all callback handlers may be expensive and error-prone.

OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_Invalid.swift
jaeopt: removed

OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_ObjcOthers.m
jaeopt: The name "testCustomNotificationCenter_Activate" looks ok to me. Not clear what you mean here.

thomaszurkan-optimizely: there is no custom notification center.  it should be testNotificationCenter_Activate

PR #229, opened 2019-06-13
use os_log for iOS/tvOS 10+

jaeopt:


OptimizelySDK/OptimizelyTests/OptimizelyTests-Common/LoggerTests.swift
```diff
@@ -38,6 +38,23 @@ class LoggerTests: XCTestCase {
     }
 }
 
+// MARK: - DefaultLogger Tests
+
+extension LoggerTests {
```
thomaszurkan-optimizely: Please add this as a test and not an extension.  Why do you have it as an extension?

PR #241, opened 2019-06-26
Code coverage improvement.

yasirfolio3:


OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_Others.swift
```diff
@@ -229,13 +229,9 @@ class OptimizelyClientTests_Others: XCTestCase {
         let exp = expectation(description: "a")
         
         optimizely.start { result in
-            switch result {
-            case .success:
-                XCTAssert(false)
-            case .failure(OptimizelyError.dataFileVersionInvalid):
+            
+            if case .failure(OptimizelyError.dataFileVersionInvalid) = result  {
```
jaeopt: With this change, we cannot catch error on .success case

OptimizelySDK/OptimizelyTests/OptimizelyTests-DataModel/ConditionLeafTests.swift
```diff
@@ -35,29 +35,30 @@ class ConditionLeafTests: XCTestCase {
     }
 
     func testDecodeSample_Invalid() {
+        
+        var tmpError: Error?
```
jaeopt: Just curious why this new change looks better to you with less readability.

OptimizelySDK/OptimizelyTests/OptimizelyTests-APIs/OptimizelyClientTests_Others.swift
yasirfolio3: Changes made.

OptimizelySDK/OptimizelyTests/OptimizelyTests-DataModel/ConditionLeafTests.swift
yasirfolio3: it always goes into one of the XCTAssert's, because of which code-cov shows the other two XCTAssert's as uncovered lines. Thats the only reason.

PR #243, opened 2019-06-27
Add Swift Package Manager (SPM) support

jaeopt:
* major changes to the project structures following SPM standard (Sources/Tests)
* tested with Xcode 11 (beta) and confirmed it works with xcode SPM ok
* tested this new structure works ok with CocoaPods as well


OptimizelySwiftSDK.podspec
```diff
@@ -12,9 +12,7 @@ Pod::Spec.new do |s|
     :git => "https://github.com/optimizely/swift-sdk.git",
     :tag => "v"+s.version.to_s
   }
-  s.source_files            = "OptimizelySDK/**/*.{swift,h}"
-  s.public_header_files     = "OptimizelySDK/OptimizelySwiftSDK/*.h"
```
thomaszurkan-optimizely: I think we still need to header files and exclude

thomaszurkan-optimizely: Can we also fix summary and homepage?
"Optimizely server-side testing core framework."
"Optimizely iOS/tvOS experiment framework" or something like that. :)

Package.swift
```diff
@@ -0,0 +1,18 @@
...
+        .tvOS(.v9)
+    ],
+    products: [
+        .library(name: "Optimizely",
+                 targets: ["Optimizely"])
+    ],
+    targets: [
+        .target(name: "Optimizely", path: "Sources")
+    ],
+    swiftLanguageVersions: [.v5]
```
thomaszurkan-optimizely: Do we want to change to use Swift 5?

OptimizelySwiftSDK.podspec
jaeopt: It looks like header files extracted from the src path by default. We can add the header line when we need support for other paths.

Package.swift
jaeopt: Yep. Package.swift is already changed to Swift 5. I'll fix others to be consistent (Swift5+).

jaeopt: I have tested with CocoaPod, Carthage, and SPM (Xcode 11). Found a couple of issues and fixed. All looks good. I'll make sure again when releasing.

PR #246, opened 2019-07-02
(WIP) add convenience method for SDK configuration

jaeopt:
* We support many configurable options and more requests coming (SSL-pinning). Currently the way of configuration support is not straightforward and spread over, so not easy to let clients use them.
* This proposal adds OptimizelyClientConfig, which collects all configurable parameters in a place.


DemoSwiftApp/AppDelegate.swift
```diff
@@ -49,7 +49,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         //     - initialize immediately with the given JSON datafile or its cached copy
         //     - no network delay, but the local copy is not guaranteed to be in sync with the server experiment settings
         
-        initializeOptimizelySDKWithCustomization()
+        //initializeOptimizelySDKWithCustomization()
```
msohailhussain: is this method deprecated?

PR #247, opened 2019-07-03
(chore):make the config a atomic property

thomaszurkan-optimizely:
After reviewing some latest use cases with Objective-C SDK. It became apparent that the developer might misuse the SDK and initialize/start multiple times.  In order to ensure that this does not cause any problems, I've made the config which is a class that holds a struct a atomic property.  This ensures that the pass by reference class is not accessed improperly even if the SDK is misused.


OptimizelySDK/Optimizely/OptimizelyClient.swift
```diff
@@ -24,7 +24,7 @@ open class OptimizelyClient: NSObject {
     // MARK: - Properties
     
     var sdkKey: String
-    var config: ProjectConfig?
+    var config: AtomicProperty<ProjectConfig> = AtomicProperty<ProjectConfig>()
```
jaeopt: It's a bit annoying to see "property" everywhere. Can we use a wrapper prop to clean this up:

var atomicConfig = AtomicProperty<ProjectConfig>()
var config: ProjectConfig? {
    get { return atomicConfig.property }
    set { atomicConfig.property = newValue }
}

PR #262, opened 2019-09-04
feat(event processor): change batching to be compliant with Event Processor

jaeopt:
* Add flush on {batchSize hit, revision change, project id change}
* Clean up batch (batch as much as possible instead of falling back to a single event mode when batch failure detected in an event group)
* Support maxQueueSize (configurable)
* Add many more unit tests to increase coverage
* Event notification is not included in this PR. It'll be prepared in a separate PR.

### DIFF
--- a/Sources/Customization/DefaultEventDispatcher.swift
+++ b/Sources/Customization/DefaultEventDispatcher.swift
@@ -66,7 +66,7 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
     }
     
     deinit {
-        removeTimer()
+        stopTimer()
 
         unsubscribe()
     }
@@ -74,7 +74,7 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
     open func dispatchEvent(event: EventForDispatch, completionHandler: DispatchCompletionHandler?) {
         dataStore.save(item: event)
         
-        setTimer()
+        startTimer()
         
         completionHandler?(.success(event.body))
     }
@@ -209,18 +209,18 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
     }
     
     func applicationDidEnterBackground() {
-        removeTimer()
+        stopTimer()
         
         flushEvents()
     }
     
     func applicationDidBecomeActive() {
         if dataStore.count > 0 {
-            setTimer()
+            startTimer()
         }
     }
     
-    func setTimer() {
+    func startTimer() {
         // timer is activated only for iOS10+ and non-zero interval value
         guard #available(iOS 10.0, tvOS 10.0, *), timerInterval > 0 else {
             flushEvents()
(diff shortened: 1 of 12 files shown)


Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -47,35 +45,79 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
...
-        timer.performAtomic { (timer) in
-            timer.invalidate()
-        }
+        stopTimer()
+
         unsubscribe()
     }
     
+    func addProjectChangeNotificationObservers() {
+        NotificationCenter.default.addObserver(forName: .didReceiveProjectIdChange, object: nil, queue: nil) { (notif) in
```
thomaszurkan-optimizely: Can we change these to didReceiveOptlyProjectIdChange?  Or something with Optimizely in there?

Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -47,35 +45,79 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
...
+            let error = OptimizelyError.eventDispatchFailed("EventQueue is full")
+            self.logger.e(error)
+            completionHandler?(.failure(error))
+            return
+        }
+        
         dataStore.save(item: event)
         
-        setTimer()
+        if dataStore.count == batchSize {
```
thomaszurkan-optimizely: greater than and equal would probably be safer

Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -88,92 +130,47 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
...
                 
-                guard let event = eventToSend else {
-                    self.logger.e(.eventBatchFailed)
-                    resetBatch()
-                    break
+                guard numEvents > 0 else { break }
+                
+                guard let batchEvent = batched else {
+                    // discard events that create invalid batch and continue
+                    removeStoredEvents(num: numEvents)
```
thomaszurkan-optimizely: i'm not sure this makes sense.  you are saying if batching failed remove all the events that were attempted to batch.

jaeopt: In fact, it discards the single event when an invalid event is detected. The comment is misleading. What happens:
- if an invalid one is found while batching, it batches all the valid ones before the invalid one and send it out.
- when trying to batch next, it finds the invalid one at the header. It discards that specific invalid one and continue batching next ones.

I'll fix the comments to avoid confusion.

thomaszurkan-optimizely: Ok, I got it.  But, numEvents should always be 1, right?  Can you call that out somehow?  You could have passed 1 into here to say remove the offending one.  But, with numEvents, its hidden unless you look at the batch function and figure out where it passes back null.

PR #264, opened 2019-09-11
feat(datafile manager): Check network reachability before downloading datafile

jaeopt:
* On asynchronous init, SDK checks if network is up. If not, it returns cached datafile immediately.
* Added more test cases for datafile download exception cases (200,304,400,500, network down)


Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -186,9 +186,9 @@ open class OptimizelyClient: NSObject {
         }
     }
     
-    func fetchDatafileBackground(resourceTimeout: Double? = nil, completion: ((OptimizelyResult<Data>) -> Void)?=nil) {
+    func fetchDatafileBackground(resourceTimeout: Double? = nil, completion: ((OptimizelyResult<Data>) -> Void)? = nil) {
         
-        datafileHandler.downloadDatafile(sdkKey: self.sdkKey, resourceTimeoutInterval: resourceTimeout) { result in
+        let processDownloadResult = { (result: OptimizelyResult<Data?>) in
```
thomaszurkan-optimizely: why?

jaeopt: Refactored for reachability support but decided to move the check to other part.

jaeopt:
This is closed for now. We’ll have a chance to review about the feature in general and its implementation (OASIS-5038)

PR #265, opened 2019-09-14
refactor: Change eventDispatch and notification to non-blocking

jaeopt:
* Fix eventDispatch to non-blocking (to avoid event serialization overhead)
* Fix all notifications to background thread
* Refactor notification logic in OptimizelyClient
* After this fix, benchmark performance (activate) shows equal or better number than ObjC


Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -366,47 +359,40 @@ open class OptimizelyClient: NSObject {
...
-        // we came from an experiment if experiment is not nil
-        if let experiment = pair?.experiment {
-            
-            var sourceInfo = [String: Any]()
-            sourceInfo[Constants.ExperimentDecisionInfoKeys.experiment] = experiment.key
-            sourceInfo[Constants.ExperimentDecisionInfoKeys.variation] = variation.key
-            decisionInfo[Constants.DecisionInfoKeys.sourceInfo] = sourceInfo
-
-            sendImpressionEvent(experiment: experiment, variation: variation, userId: userId, attributes: attributes)
+        let experiment = pair?.experiment
```
thomaszurkan-optimizely: this line makes no sense.

PR #270, opened 2019-10-16
fix: add close() API to OptimizelyClient and EventDispatcher for test support

jaeopt:
- async event dispatch and notification makes testing complicated (timing control)
- add an extra API call: close() for testing support only
- it stops datafile polling and flush all events synchronously

### DIFF
--- a/DemoSwiftApp/AppDelegate.swift
+++ b/DemoSwiftApp/AppDelegate.swift
@@ -152,7 +152,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             }
         })
         
-        _ = optimizely.notificationCenter.addLogEventNotificationListener(logEventListener: { (url, event) in
+        _ = notificationCenter.addLogEventNotificationListener(logEventListener: { (url, event) in
             print("Received logEvent notification: \(url) \(event)")
         })
     }
--- a/Sources/Customization/DefaultEventDispatcher.swift
+++ b/Sources/Customization/DefaultEventDispatcher.swift
@@ -276,7 +276,7 @@ extension DefaultEventDispatcher {
     
     // MARK: - Tests
 
-    func close() {
+    open func close() {
         self.flushEvents()
         self.dispatcher.sync {}
     }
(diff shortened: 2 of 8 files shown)


DemoSwiftApp/AppDelegate.swift
```diff
@@ -152,7 +152,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             }
         })
         
-        _ = optimizely.notificationCenter.addLogEventNotificationListener(logEventListener: { (url, event) in
+        _ = notificationCenter.addLogEventNotificationListener(logEventListener: { (url, event) in
```
thomaszurkan-optimizely: fixed in another PR

Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -276,4 +274,11 @@ extension DefaultEventDispatcher {
         }
     }
     
+    // MARK: - Tests
+
+    open func close() {
```
thomaszurkan-optimizely: should this be open?

Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -256,12 +254,12 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
 extension DefaultEventDispatcher {
     
     func addProjectChangeNotificationObservers() {
-        observerProjectId = NotificationCenter.default.addObserver(forName: .didReceiveOptimizelyProjectIdChange, object: nil, queue: nil) { [weak self] (notif) in
+        observerProjectId = NotificationCenter.default.addObserver(forName: .didReceiveOptimizelyProjectIdChange, object: nil, queue: nil) { [weak self] (_) in
```
thomaszurkan-optimizely: This would never happen since the project config is based on the client/ sdk key.

jaeopt: A good catch. I see it's never called in the current implementation of ProjectConfig. We'll keep this as is and we'll see if we'll have the usage case later.

PR #272, opened 2019-10-23
fix: change EventDispatcher to EventProcessor + EventHandler

jaeopt:


Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+        
+        dataStore.save(item: EventForDispatch(body: body))
+        
+        if dataStore.count >= batchSize {
+            flush()
+        } else {
+            startTimer()
+        }
+        
+        completionHandler?(.success(body))
```
mjc1283: Does completion handler make sense here? Nothing is asynchronous.

Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+* limitations under the License.                                           *
+***************************************************************************/
+
+import Foundation
+
+public enum DataStoreType {
+    case file, memory, userDefaults
+}
+
+open class DefaultEventProcessor: BackgroundingCallbacks, OPTEventProcessor {
```
mikeng13: Let's call this `BatchEventProcessor` since the word `Default` doesn't say much about it's implementation

Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+    func startTimer() {
+        // timer is activated only for iOS10+ and non-zero interval value
+        guard #available(iOS 10.0, tvOS 10.0, *), timerInterval > 0 else {
+            flush()
+            return
+        }
+        
+        guard self.timer.property == nil else { return }
+        
+        DispatchQueue.main.async {
```
mjc1283: Why do we use `DispatchQueue.main` for the timer task, but our own private `DispatchQueue` for flushing?

Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+    // timer-interval for batching (0 = no batching, negative = use default)
+    var timerInterval: TimeInterval
+    // batch size (1 = no batching, 0 or negative = use default)
+    // attempt to send events in batches with batchSize number of events combined
+    var batchSize: Int
+    var maxQueueSize: Int
+    
+    public struct DefaultValues {
+        static public let batchSize = 10
+        static public let timeInterval: TimeInterval = 60  // secs
```
mikeng13: The default batch interval in other SDKs is 30 secs

Sources/Customization/Protocols/OPTEventProcessor.swift
```diff
@@ -0,0 +1,34 @@
...
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+
+import Foundation
+
+public protocol OPTEventProcessor {
+    
+    func process(event: BatchEvent, completionHandler: DispatchCompletionHandler?)
```
mikeng13: We should be processing `UserEvent`s such as `ImpressionEvent` or `ConversionEvent`. Please take a look at these: 

https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/event/internal/UserEvent.java

https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/event/internal/ConversionEvent.java

The "batch" should be a batch created from `UserEvent`s

Sources/Customization/DefaultEventProcessor.swift
jaeopt: This was discussed before. Default interval should be controlled separately per SDK since it should depend on applications.

Sources/Customization/Protocols/OPTEventProcessor.swift
jaeopt: I understand BatchEvent is a wrong name here, but we used it for a while. I thought about changing the name and create new UserEvent type, but we already have too much name/interface changes. Will look at possible changes.

Sources/Customization/DefaultEventProcessor.swift
thomaszurkan-optimizely: in that async call you are setting a timer which is part of the main event loop.  thus it must be done on the main thread.  we had considered moving this to a dispatch timer but felt this was right for background foreground.  if you attempt a timer call in the background it will crash as backgrounding takes a delegate pattern as opposed to callback.

Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+    deinit {
+        stopTimer()
+        
+        removeProjectChangeNotificationObservers()
+
+        unsubscribe()
+    }
+    
+    open func process(event: BatchEvent, completionHandler: DispatchCompletionHandler? = nil) {
+        guard dataStore.count < maxQueueSize else {
```
thomaszurkan-optimizely: no flush?

Sources/Customization/DefaultEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+            let error = OptimizelyError.eventDispatchFailed("EventQueue is full")
+            self.logger.e(error)
+            completionHandler?(.failure(error))
+            return
+        }
+        
+        guard let body = try? JSONEncoder().encode(event) else {
```
thomaszurkan-optimizely: this is now on the main thread if you are calling process from event processor instead of handler.

Sources/Customization/DefaultEventProcessor.swift
jaeopt: In fact, it's asynchronous. We serialize the event to be stored persistently, which takes a significant delay. That's why we call processEvent asynchronously. We have a couple of cases for process failures, so handler callback will be useful. It's optional.

Sources/Customization/DefaultEventProcessor.swift
mikeng13: Ah yeah you are right.

Sources/Customization/Protocols/OPTEventProcessor.swift
mikeng13: But this is a new protocol right? We shouldn't be worries about `BatchEvent` breaking a new protocol since no one is using it yet

Sources/Customization/DefaultEventProcessor.swift
jaeopt: Not clear what you mean here. That line is for overflow dropping and flush on batch size is handled below. Can you clarify if you mean something else?

Sources/Customization/DefaultEventProcessor.swift
jaeopt: It's called asynchronously on background thread from OptimizelyClient. Check out OptimizelyClient.

Sources/Customization/Protocols/OPTEventProcessor.swift
jaeopt: Yep, it's not about breaking changes. I'll fix it.

Sources/Customization/BatchEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+            self.dataStore = DataStoreQueueStackImpl<EventForDispatch>(queueStackName: "OPTEventQueue",
+                                                                       dataStore: DataStoreUserDefaults())
+        }
+        
+        if self.maxQueueSize < self.batchSize {
+            self.logger.e(.eventDispatcherConfigError("batchSize cannot be bigger than maxQueueSize"))
+            self.maxQueueSize = self.batchSize
+        }
+        
+        addProjectChangeNotificationObservers()
```
thomaszurkan-optimizely: I'd like to remove this and use the notification center as a property that is injected.

Sources/Customization/Protocols/OPTEventHandler.swift
```diff
@@ -0,0 +1,29 @@
...
+* Unless required by applicable law or agreed to in writing, software      *
+* distributed under the License is distributed on an "AS IS" BASIS,        *
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+import Foundation
+
+public protocol OPTEventHandler {
```
thomaszurkan-optimizely: I'd like to discuss.  We want to deprecate DefaultEventDispatcher and explain that dispatch now just sends and is part of event dispatcher.  Then, we create a new EventDispatcher that is HttpEventProcessor.

Sources/Customization/Protocols/OPTEventProcessor.swift
thomaszurkan-optimizely: whispering.... we can do a typealias to userevent = batch event.

jaeopt: Not a bad idea, but I already made all the changes to support UserEvent :)

DemoSwiftApp/AppDelegate.swift
```diff
@@ -99,9 +99,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // 30 sec interval may be too frequent. This is for demo purpose.
         // This should be should be much larger (default = 10 mins).
         let customDownloadIntervalInSecs = 30
+        
+        let eventProcessor = BatchEventProcessor(batchSize: 1)
```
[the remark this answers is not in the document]
jaeopt: Will fix it more realistic number. Set for testing for now.

Sources/Customization/BatchEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+                
+                // we've exhuasted our failure count.  Give up and try the next time a event
+                // is queued or someone calls flush.
+                if failureCount > DefaultValues.maxFailureCount {
+                    self.logger.e(.eventSendRetyFailed(failureCount))
+                    break
+                }
+                
+                // send notification BEFORE sending event to the server
+                NotificationCenter.default.post(name: .willSendOptimizelyEvents, object: batchEvent)
```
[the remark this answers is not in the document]
jaeopt: Will take a look at it. An issue is that we do not have sdkKey to inject the right NotificationCenter since EventProcessor is global, shared for all sdkKeys.

Sources/Customization/BatchEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+        guard dataStore.count < maxQueueSize else {
+            let error = OptimizelyError.eventDispatchFailed("EventQueue is full")
+            self.logger.e(error)
+            completionHandler?(.failure(error))
+            return
+        }
+        
+        guard let body = try? JSONEncoder().encode(event) else {
+            let error = OptimizelyError.eventDispatchFailed("Event serialization failed")
+            self.logger.e(error)
```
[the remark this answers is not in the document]
jaeopt: I do not think it's necessary here. Before the queue is full, we expect flush tried many times (whenever it hits or over the batchSize). So, I expect the queue overflows only when previous flushes did not work effectively (like network down, etc). In that case, another flush before dropping will not help.

Sources/Customization/BatchEventProcessor.swift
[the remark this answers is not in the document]
thomaszurkan-optimizely: That is a good point.  But, this breaks down as well.  What if you have multiple clients running?  Wouldn't they all be notified?  In general the notification center is not a great mechanism when working on a large team.  In previous positions, we discovered that debugging became much more difficult using notification center.  Support for multiple delegates to be notified worked much better and was clearer when trying to find delegate implementations.

Sources/Customization/BatchEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+    
+    open func process(event: UserEvent, completionHandler: DispatchCompletionHandler? = nil) {
+        guard dataStore.count < maxQueueSize else {
+            let error = OptimizelyError.eventDispatchFailed("EventQueue is full")
+            self.logger.e(error)
+            completionHandler?(.failure(error))
+            return
+        }
+        
+        guard let body = try? JSONEncoder().encode(event.batchEvent) else {
```
thomaszurkan-optimizely: Let's wrap the encoding and saving in a Dispatch.global().async call.

Sources/Customization/BatchEventProcessor.swift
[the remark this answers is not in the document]
thomaszurkan-optimizely: I still disagree because we don't know what the use of the phone is.  We also have no guarantee that the queue size and batch size have been set optimally.

Sources/Customization/BatchEventProcessor.swift
```diff
@@ -0,0 +1,271 @@
...
+    }
+    
+    func applicationDidEnterBackground() {
+        stopTimer()
+        
+        flush()
+    }
+    
+    func applicationDidBecomeActive() {
+        if dataStore.count > 0 {
```
thomaszurkan-optimizely: change to flush?

Sources/Customization/Protocols/OPTEventHandler.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    /// Dispatch event to Optimizely backend for results measurement.
+    ///
+    /// - Parameters:
+    ///   - event: EventForDispatch object which contains the url to send to and the body.
+    ///   - completionHandler: Called when the event has been sent or if an error occured.
+    ///     This may not be called in the case where the dispatcher is doing batch events. It is up to the implementor of the protocol.
+    func dispatch(event: EventForDispatch, completionHandler: DispatchCompletionHandler?)
```
thomaszurkan-optimizely: Can we just replace the EventHandler with EventDispatcherV2?

Sources/Customization/Protocols/OPTEventProcessor.swift
```diff
@@ -0,0 +1,34 @@
...
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+
+import Foundation
+
+public protocol OPTEventProcessor {
+    
+    func process(event: UserEvent, completionHandler: DispatchCompletionHandler?)
```
thomaszurkan-optimizely: Create a ProcessorCompletionHandler (it can just take the user event and that way you can do the json encoding async in another thread.

Sources/Optimizely/OptimizelyClient+ObjC.swift
```diff
@@ -304,13 +322,84 @@ extension OptimizelyClient {
         try self.track(eventKey: eventKey, userId: userId, attributes: attributes, eventTags: eventTags)
     }
     
+    @available(swift, obsoleted: 1.0)
```
thomaszurkan-optimizely: We need to talk about this one.  I don't feel comfortable making something like this public.  Too much chance for error.

PR #273, opened 2019-11-05
add/fix tests for EP+ED refactor

jaeopt:
- add new tests for EP+ED new architecture
- move old tests for EP to a legacy test group
- fix demo apps for new EP+ED 
- cleanup SamplesForAPI codes


jaeopt:
Will be reviewed when next major version is released since this requires API changes.

PR #281, opened 2020-01-09
wip: use subscripting to always access latest

thomaszurkan-optimizely:


thomaszurkan-optimizely:
- I brought it up in the devx channel because I wanted feedback from others on whether we should do this for all sdks. 

- Limitations for expected common usage models. For example, how to get all keys, how to get all experiments/features. Config-sync overhead for each item access and consistency issue (array data mixed from two different datafiles) I'm also concerned too.

Please expand on "Config-sync overhead for each item access and consistency issue".  I don't follow you there.  

How are you proposing that users keep track of experiments and features now?  They get a map that they have to access in the same manner as the container subscript protocols.  It just makes it real obvious that you shouldn't hold onto the map/struct.  Of course, we also need to mention this in documentation.  

Are you just planning on using the datafile change notification to tell the user that their Optimizely config is stale? Is that enough?

jaeopt:
I think it all depends on expected usage models. We can move the discussion to Slack to share with the group. 
If you propose the change for all SDKs, I think it's better to close the current implementation as is and we can discuss with the entire group your proposal for next phase (API change proposal).

PR #282, opened 2020-01-13
chore: prepare for release 3.2.0

jaeopt:


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+    * Call `getOptimizelyConfig()` to get a snapshot copy of project configuration static data.
+    * It returns an `OptimizelyConfig` instance which includes a datafile revision number, all experiments, and feature flags mapped by their key values.
+    * For details, refer to a documention page: https://docs.developers.optimizely.com/full-stack/docs/optimizelyconfig-swift
+* Add LogEvent Notification ([#263](https://github.com/optimizely/swift-sdk/pull/263)):
+    * Register an event listner with `addLogEventNotificationListener()`
+    * The lister will be called when events are dispatched to the server.
+* Turn on "Allow app extensions only" flag in build settings ([#277](https://github.com/optimizely/swift-sdk/pull/277))
+
+### Bug Fixes
+- When timerInterval is set to a negative value, EventDispatcher is changed to use the default interval value for batching (instead of disabling batch). Batching is  disabled when the timerInterval is set to zero.  ([#268](https://github.com/optimizely/swift-sdk/pull/268))
```
thomaszurkan-optimizely: Is this right?  It doesn't sound the like the right approach.  Batching should be off by setting the batch size to one.

jaeopt: It has been changed to be consistent with EP spec. That's the way FSC expects.

jaeopt: The test failures are due to flakiness which has been fixed in EP+ED changes before, but which has been reverted. This is not related to SDK issues. I'll rerun and make sure all tests pass ok. The correct fix for the flakiness will be brought back in later.

PR #296, opened 2020-02-12
refact: clean up SDK init methods with polling control

jaeopt:
* SDK init with and without polling are controlled with separate init methods, which conflict with each other. Cleaned up.
* Improved SDK initialization samples


DemoSwiftApp/AppDelegate.swift
```diff
@@ -95,17 +95,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func initializeOptimizelySDKWithCustomization() {
         // customization example (optional)
 
-        let customLogger = CustomLogger()
-        // 60 sec interval may be too frequent. This is for demo purpose.
-        // This should be should be much larger (default = 10 mins).
-        let customDownloadIntervalInSecs = 60
+        // You can enable background datafile polling by setting periodicDownloadInterval (polling is disabled by default)
```
thomaszurkan-optimizely: nit: foreground datafile polling

PR #297, opened 2020-02-14
fix: sync init flow changed to support dynamic update

jaeopt:
* Add an option to dynamically update ProjectConfig on datafile cache update for sync init flow (this feature is disabled by default to be compatible with the old versions).
* When periodic polling is enabled, the sync cache change always update ProjectConfig dynamically regardless of the setting above.

### DIFF
--- a/Sources/Optimizely/OptimizelyClient.swift
+++ b/Sources/Optimizely/OptimizelyClient.swift
@@ -24,6 +24,8 @@ open class OptimizelyClient: NSObject {
     // MARK: - Properties
     
     var sdkKey: String
+    var periodicDownloadInterval = 0
+    
     private var atomicConfig: AtomicProperty<ProjectConfig> = AtomicProperty<ProjectConfig>()
     var config: ProjectConfig? {
         get {
@@ -80,6 +82,7 @@ open class OptimizelyClient: NSObject {
                 defaultLogLevel: OptimizelyLogLevel? = nil) {
         
         self.sdkKey = sdkKey
+        self.periodicDownloadInterval = 0  // this will be updated with the new API change
         
         super.init()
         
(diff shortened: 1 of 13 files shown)


Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -139,80 +149,74 @@ open class OptimizelyClient: NSObject {
     ///   - datafile: This datafile will be used when cached copy is not available (fresh start)
     ///             A cached copy from previous download is used if it's available.
     ///             The datafile will be updated from the server in the background thread.
+    ///   - doUpdateConfigOnNewDatafile: When a new datafile is fetched from the server in the background thread, the SDK will be updated with the new datafile immediately if this value is set to true. When it's set to false (default), the new datafile is cached and will be used when the SDK is started again.
     ///   - doFetchDatafileBackground: This is for debugging purposes when
     ///             you don't want to download the datafile.  In practice, you should allow the
     ///             background thread to update the cache copy (optional)
-    public func start(datafile: Data, doFetchDatafileBackground: Bool = true) throws {
+    public func start(datafile: Data,
+                      doUpdateConfigOnNewDatafile: Bool = false,
```
thomaszurkan-optimizely: These two don't really make sense here.  If you have doUpdateConfigOnNewDatafile but you have doFeatchDatafileBackground as false.  It is an invalid combination.  Maybe an enum might be better here with all valid combinations.  Also, the default combination is probably not ideal if you have polling enabled which is not addressed in this combo or PR.  So, you might want combos that include polling and what to do there.

jaeopt: Yes it's not so clean as it should be. The reason I keep "doFetchDatafileBackground" is for API backward compatibility only, where we had this option for debugging support. We can let them override "doUpdateConfigOnNewDatafile" when they still want to disable fetch completely for debugging purpose.

Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -39,6 +40,14 @@ open class OptimizelyClient: NSObject {
     }
     
     let eventLock = DispatchQueue(label: "com.optimizely.client")
+    
+    private var isPeriodicPollingEnabled: Bool {
```
thomaszurkan-optimizely: nit: can we make this an extension?

PR #301, opened 2020-03-02
fix: remove accumulated polling when app comes back to foreground

jaeopt:
* When app goes to background, polling stops OK as expected.
* But when app comes back to foreground, the current implementation looks at the timer behind and try to poll multiple times, which is annoying. This fix cleans up this logic, so recover normal periodic polling regardless of delayed timer value.

### DIFF
--- a/Sources/Implementation/DefaultDatafileHandler.swift
+++ b/Sources/Implementation/DefaultDatafileHandler.swift
@@ -1,18 +1,18 @@
 /****************************************************************************
-* Copyright 2019, Optimizely, Inc. and contributors                        *
-*                                                                          *
-* Licensed under the Apache License, Version 2.0 (the "License");          *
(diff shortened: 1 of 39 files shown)


Sources/Implementation/DefaultDatafileHandler.swift
```diff
@@ -159,19 +159,6 @@ class DefaultDatafileHandler: OPTDatafileHandler {
         }
     }
     
-    @objc
```
thomaszurkan-optimizely: why was this removed?

jaeopt: It is not used anywhere. probably leftover from previous changes.

thomaszurkan-optimizely: leftover from ios 9 support. got it. :)

PR #304, opened 2020-03-09
(fix): show alternative using atomic property for thread safe notification center

thomaszurkan-optimizely:


Sources/Implementation/DefaultNotificationCenter.swift
```diff
@@ -134,26 +161,34 @@ public class DefaultNotificationCenter: OPTNotificationCenter {
...
+        _listeners.performAtomic { (listeners) in
+            listeners.notificationListeners.removeAll()
+        }
     }
     
     public func sendNotifications(type: Int, args: [Any?]) {
-        for values in notificationListeners.values where values.0 == type {
-            values.1(args)
+        _listeners.performAtomic { (listeners) in
+            for values in listeners.notificationListeners.values where values.0 == type {
```
jaeopt: Holding all listener calls in the atomic for a while won't be safe. Let's grap the listeners synchronously and call them outside the atomic block.

Sources/Implementation/DefaultNotificationCenter.swift
```diff
@@ -17,33 +17,60 @@
 import Foundation
 
 public class DefaultNotificationCenter: OPTNotificationCenter {
-    public var notificationId: Int = 1
-    var notificationListeners = [Int: (Int, GenericListener)]()
+    public var notificationId: Int {
+        get {
+            return notifications.notificationId
```
jaeopt: We need sync-er for allowing access to notificationId (public), which can cause thread conflict with incrementNotificationId.

Sources/Implementation/DefaultNotificationCenter.swift
thomaszurkan-optimizely: i agree with that if there is no problem with a removed listener possibly getting one more notify which could happen if you are iterating through a copy of the list.

Sources/Implementation/DefaultNotificationCenter.swift
thomaszurkan-optimizely: i would prefer to remove this access all together.  it doesn't make sense.

jaeopt: Yep, this should be removed, but it was open to public. Lets keep it for now and clean it up on the next major release.

PR #310, opened 2020-03-20
chore (OPTDataStore): Remove print statements and change memory store default backing store.

thomaszurkan-optimizely:
## Summary
1. remove all print statements. 
2. datastore memory backing store configurable.
3. datastore memory backing changed from file to user defaults.
4. default event handler uses memory with user defaults for tvOS and file based for iOS.
5. max bytes for tvOS User Defaults datastore is 128k. 1M for iOS.  Added unit tests.
6. changed datafile handler to use data store file.
7. fixed unit tests and tweaked data store file.  Added removeItem to OPTDataStore.


Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
+                get {
+                    return OptimizelyLogLevel.info
+                }
+            }
+            
+            func log(level: OptimizelyLogLevel, message: String) {
+                messages.append(message)
+            }
+            
+
```
jaeopt: remove spaces

Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
+        class Logger : OPTLogger {
+            public var messages: [String]
+                = []
+            required init() {
+                
+            }
+            static var logLevel: OptimizelyLogLevel {
+                set {
+                    
+                }
```
jaeopt: remove

Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
         
-        XCTAssert(v2 == value)
+        var array = [Data]()
+        for _ in 0 ... DataStoreUserDefaults.MAX_DS_SIZE + 1 {
+            array.append("01234567890abcdef".data(using: .ascii)!)
+        }
+        
+        datastore.saveItem(forKey: "testUserDefaultsTooBig", value: array)
+        
+        let value = datastore.getItem(forKey: "testUserDefaultsTooBig") as? String
```
jaeopt:
```suggestion
        let value = datastore.getItem(forKey: "testUserDefaultsTooBig") as? [Data]
```

Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
+        XCTAssert(value == nil)
+        XCTAssert(logger.messages.last!.contains("Save to User Defaults error: testUserDefaultsTooBig is too big to save size"))
+        HandlerRegistryService.shared.binders.property?.removeAll()
 
     }
-
     func testPerformanceExample() {
         // This is an example of a performance test case.
         self.measure {
```
jaeopt: remove

Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
         let value = datastore.getItem(forKey: "testString") as! String
         
         XCTAssert(value == "value")
         
-        datastore.unsubscribe()
+    }
+
+    func testUserDefaultsTooBig() {
+        // This is an example of a functional test case.
+        // Use XCTAssert and related functions to verify your tests produce the correct results.
```
jaeopt: remove

Tests/OptimizelyTests-Common/DataStoreTests.swift
```diff
@@ -97,26 +97,64 @@ class DataStoreTests: XCTestCase {
...
+        let logger = Logger()
         
-        let v2 = datastore.getItem(forKey: "testString") as! String
+        let binder: Binder = Binder<OPTLogger>(service: OPTLogger.self).to { () -> OPTLogger? in
+            return logger
+        }
+
+        HandlerRegistryService.shared.registerBinding(binder: binder)
         
-        XCTAssert(v2 == value)
```
jaeopt: Let's clean up here test data which may be left from previous test:
   UserDefaults.standard.removeObject(forKey: "testUserDefaultsTooBig")

thomaszurkan-optimizely: it is never saved because it is too big.  :)

Sources/Implementation/DefaultDatafileHandler.swift
```diff
@@ -255,56 +257,57 @@ class DefaultDatafileHandler: OPTDatafileHandler {
...
+extension OPTDataStore {
+    public func removeItem(sdkKey:String) {
+        if let store = self as? DataStoreUserDefaults {
+            store.removeItem(sdkKey: sdkKey)
+        }
+        else if let store = self as? DataStoreFile<Data> {
+            store.removeItem(sdkKey: sdkKey)
+        }
+        else if let store = self as? DataStoreMemory<Data> {
+            store.removeItem(sdkKey: sdkKey)
```
jaeopt: It must be for backward compatibility. Can't we leave it empty? All our datastores override it. Also we can move it to the end of OPTDataStore for readability.

public func removeItem(sdkKey: String) { }

Sources/Implementation/DefaultDatafileHandler.swift
```diff
@@ -255,56 +257,57 @@ class DefaultDatafileHandler: OPTDatafileHandler {
...
-            } catch {/* error handling here */
-                logger.e("Problem saving datafile for key " + sdkKey)
-            }
+    func getDatafileCache(sdkKey:String) -> OPTDataStore {
+        if let cache = datafileCache[sdkKey] {
+            return cache
+        }
+        else {
+            #if os(tvOS)
+            let store = DataStoreUserDefaults()
```
jaeopt: It looks good, but I'm concerned about the datafile size which can grow big more than MB. They may not fit into UserDefaults max configs in many cases. Also with event queues, it can take substantial spaces in UserDefaults even if they can fit in 128KB each.

thomaszurkan-optimizely: even if you reach 512k in tvOS, the action is that the user is notified, it can actually grow beyond that.  So, this is probably pretty safe.  That said, if the datafile is too big, what choice do they have on tvOS?

Sources/Implementation/DefaultDatafileHandler.swift
thomaszurkan-optimizely: no because that is the method called on the extension (i.e. the other implementation is not an override since it is an extension).  the only alternative is to add it to the data store protocol directly.

Sources/Implementation/DefaultDatafileHandler.swift
thomaszurkan-optimizely: basically, the same thing will happen as happens now on tvOS.  start async, caching fails, but you can continue.  the bigger question may be do we want to disable polling for tvOS?  i am really curious why the simulator doesn't fail.

jaeopt: Polling is optional, so we don't need to block it. They won't turn it on if not working. 
We can also consider "Cache" folder as done in ObjC. Both methods are not fully reliable for keeping datafile cache and events.

PR #315, opened 2020-04-15
feat: implementing OptimizelyJSON

yasirfolio3:
# Summary
- implementing OptimizelyJSON object and its API's needed to support json access for new public SDK API's.

### DIFF
--- a/Sources/Data Model/OptimizelyJSON.swift
+++ b/Sources/Data Model/OptimizelyJSON.swift
@@ -20,7 +20,7 @@ class OptimizelyJSON {
     
     private typealias schemaHandler = (Any) throws -> Void
     private var payload: String?
-    private var data: [String: Any]?
+    private var data = [String: Any]()
     
     // MARK: - Init
     
@@ -29,7 +29,7 @@ class OptimizelyJSON {
             guard let data = payload.data(using: .utf8),
                 let jsonData = try JSONSerialization.jsonObject(with: data, options : .allowFragments) as? [String: Any] else
             {
-                throw OptimizelyError.invalidJSONString
+                throw OptimizelyError.failedToConvertStringToDictionary
             }
             self.data = jsonData
             self.payload = payload
@@ -38,7 +38,10 @@ class OptimizelyJSON {
         }
     }
     
-    init(data: [String: Any]) {
+    init(data: [String: Any]) throws {
+        if !JSONSerialization.isValidJSONObject(data) {
+            throw OptimizelyError.invalidDictionary
+        }
         self.data = data
     }
     
@@ -50,19 +53,15 @@ class OptimizelyJSON {
     /// - Throws: `OptimizelyError`
     func toString() throws -> String {
         guard let payload = self.payload else {
-            guard let data = self.data, JSONSerialization.isValidJSONObject(data) else {
-                throw OptimizelyError.invalidJSONDictionary
+            do {
+                let jsonData = try JSONSerialization.data(withJSONObject: data, options: [])
+                if let jsonString = String(data: jsonData, encoding: .utf8) {
+                    self.payload = jsonString
+                    return jsonString
+                }
+            } catch {
             }
-            guard let jsonData = try? JSONSerialization.data(
-                withJSONObject: data,
-                options: .prettyPrinted
-                ),
-                let jsonString = String(data: jsonData,
-                                        encoding: String.Encoding.ascii) else {
-                                            throw OptimizelyError.failedToConvertJSONToString
-            }
-            self.payload = jsonString
-            return jsonString
+            throw OptimizelyError.failedToConvertDictionaryToString
         }
         return payload
     }
@@ -70,39 +69,42 @@ class OptimizelyJSON {
     /// - Returns:  the json dictionary
     /// - Throws: `OptimizelyError`
     func toMap() throws -> [String: Any] {
-        guard let data = self.data else {
-            throw OptimizelyError.invalidJSONDictionary
-        }
         return data
     }
     
     /// populates the decodable schema passed by the user
     /// - Parameters:
     ///   - jsonPath: Key path for the value.
-    ///   - decodableSchema: Schema to decode into.
+    ///   - schema: Decodable schema to populate.
     /// - Throws: `OptimizelyError`
-    func getValue<T: Decodable>(jsonPath: String, decodableSchema: inout T) throws {
+    func getValue<T: Decodable>(jsonPath: String, schema: inout T) throws {
         func populateDecodableSchema(value: Any) throws {
-            guard JSONSerialization.isValidJSONObject(value) else {
-                throw OptimizelyError.invalidJSONDictionary
+            // Try to decode value into schema
+            if JSONSerialization.isValidJSONObject(value) {
+                do {
+                    let jsonData = try JSONSerialization.data(
+                        withJSONObject: value,
+                        options: .prettyPrinted
+                    )
+                    schema = try JSONDecoder().decode(T.self, from: jsonData)
+                    return
+                } catch {
+                }
             }
-            do {
-                let jsonData = try JSONSerialization.data(
-                    withJSONObject: value,
-                    options: .prettyPrinted
-                )
-                decodableSchema = try JSONDecoder().decode(T.self, from: jsonData)
-            } catch {
-                throw OptimizelyError.failedToAssignValueToSchema
+            // Assign directly to schema if possible
+            if let v = value as? T {
+                schema = v
+                return
             }
+            throw OptimizelyError.failedToAssignValueToSchema
         }
-        try findValue(jsonPath: jsonPath, schemaHandler: populateDecodableSchema(value:))
+        try getValue(jsonPath: jsonPath, schemaHandler: populateDecodableSchema(value:))
     }
     
     /// populates the schema passed by the user
     /// - Parameters:
     ///   - jsonPath: Key path for the value.
-    ///   - decodableSchema: Schema to decode into.
+    ///   - schema: schema to populate.
     /// - Throws: `OptimizelyError`
     func getValue<T>(jsonPath: String, schema: inout T) throws {
         func populateSchema(value: Any) throws {
@@ -112,24 +114,19 @@ class OptimizelyJSON {
             }
             throw OptimizelyError.failedToAssignValueToSchema
         }
-        try findValue(jsonPath: jsonPath, schemaHandler: populateSchema(value:))
+        try getValue(jsonPath: jsonPath, schemaHandler: populateSchema(value:))
     }
     
-    private func findValue(jsonPath: String, schemaHandler: schemaHandler) throws {
-        let pathArray = jsonPath.components(separatedBy: ".")
-        if pathArray.count == 0 {
+    private func getValue(jsonPath: String, schemaHandler: schemaHandler) throws {
+        
+        if jsonPath == "" {
             // Populate the whole schema
-            guard let data = self.data, JSONSerialization.isValidJSONObject(data) else {
-                throw OptimizelyError.invalidJSONDictionary
-            }
             try schemaHandler(data)
             return
         }
         
+        let pathArray = jsonPath.components(separatedBy: ".")
         let lastIndex = pathArray.count - 1
-        guard let data = self.data else {
-            throw OptimizelyError.invalidJSONDictionary
-        }
         
         var internalMap = data
         for (index,key) in pathArray.enumerated() {
--- a/Sources/Data Model/OptimizelyJSON.swift
+++ b/Sources/Data Model/OptimizelyJSON.swift
@@ -1,0 +1,147 @@
+/****************************************************************************
+* Copyright 2020, Optimizely, Inc. and contributors                        *
+*                                                                          *
+* Licensed under the Apache License, Version 2.0 (the "License");          *
+* you may not use this file except in compliance with the License.         *
+* You may obtain a copy of the License at                                  *
+*                                                                          *
+*    http://www.apache.org/licenses/LICENSE-2.0                            *
+*                                                                          *
+* Unless required by applicable law or agreed to in writing, software      *
+* distributed under the License is distributed on an "AS IS" BASIS,        *
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+* See the License for the specific language governing permissions and      *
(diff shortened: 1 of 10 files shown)


Sources/Data Model/OptimizelyJSON.swift
```diff
@@ -0,0 +1,144 @@
+/****************************************************************************
```
msohailhussain, line 1: it's not a data model, it should be moved.

Sources/Data Model/OptimizelyJSON.swift
```diff
@@ -0,0 +1,144 @@
...
+    
+    /// populates the decodable schema passed by the user
+    /// - Parameters:
+    ///   - jsonPath: Key path for the value.
+    ///   - schema: Decodable schema to populate.
+    /// - Throws: `OptimizelyError`
+    func getValue<T: Decodable>(jsonPath: String, schema: inout T) throws {
+        func populateDecodableSchema(value: Any) throws {
+            // Try to decode value into schema
+            if JSONSerialization.isValidJSONObject(value) {
```
msohailhussain, line 83: Shouldn't we use `guard` if that's not a valid JSON object.

Sources/Data Model/OptimizelyJSON.swift
```diff
@@ -0,0 +1,144 @@
...
+                do {
+                    let jsonData = try JSONSerialization.data(
+                        withJSONObject: value,
+                        options: .prettyPrinted
```
msohailhussain, line 87: Do we need this option, after all, user will get everything in object. not as string.

Sources/Data Model/OptimizelyJSON.swift
```diff
@@ -0,0 +1,144 @@
...
+                    )
+                    schema = try JSONDecoder().decode(T.self, from: jsonData)
+                    return
+                } catch {
```
msohailhussain, line 91: Why we are swallowing error?

Sources/Data Model/OptimizelyJSON.swift
```diff
@@ -0,0 +1,144 @@
...
+        }
+        try getValue(jsonPath: jsonPath, schemaHandler: populateDecodableSchema(value:))
+    }
+    
+    /// populates the schema passed by the user
+    /// - Parameters:
+    ///   - jsonPath: Key path for the value.
+    ///   - schema: schema to populate.
+    /// - Throws: `OptimizelyError`
+    func getValue<T>(jsonPath: String, schema: inout T) throws {
```
msohailhussain, line 109: what's the role of this method?

Sources/Data Model/OptimizelyJSON.swift
msohailhussain: Also access modifiers should be public? Question for @pawels-optimizely ?

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+import XCTest
+
+// MARK: - Sample Data and Setup
+
+class OptimizelyJSONTests: XCTestCase {
+    
+    private struct ValidSchema: Decodable, Equatable {
```
msohailhussain, line 23: I think sequence should be `method`, `enum coding keys` then `variables

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+                return
+            }
+            throw DecodingError.typeMismatch(InnerField2.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Wrong type for InnerField2"))
+        }
+    }
+    
+    private struct EmptySchema: Decodable {
+    }
+    
+    private struct IncorrectSchema: Decodable {
```
msohailhussain, line 85: I'll recommend add one more schema, that should be used directly with `getValue<T>` without encodable.

Sources/Data Model/OptimizelyJSON.swift
pawels-optimizely: not sure how it is in c#, but usually you need a public access to an object to return from a different package, so I would say it should be public, unless you are asking about something else.

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+        ]
+        try! self.optimizelyJSON = OptimizelyJSON(payload: self.payload)
+    }
+}
+
+// MARK: - Tests
+
+extension OptimizelyJSONTests {
+    
+    func testConstructorsHappyPath() {
```
msohailhussain, line 129: Please break down into multiple cases. it's a long test case verifying multiple things.

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+            try optimizelyJSON = OptimizelyJSON(payload: "incorrect_string")
+        } catch {
+            err = error
+        }
+        XCTAssertEqual(err?.localizedDescription, OptimizelyError.failedToConvertStringToDictionary.reason)
+        XCTAssertNil(optimizelyJSON)
+    }
+    
+    func testConstructorWithInvalidJSONArrayPayloadString() {
+        let jsonArrayPayload = """
```
msohailhussain, line 193: what's wrong here?

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+    
+    func testGetValueForInvalidJSONKeyAndEmptySchema() {
+        var err: Error?
+        var schema = EmptySchema()
+        do {
+            try self.optimizelyJSON.getValue(jsonPath: "field4.", schema: &schema)
+        } catch {
+            err = error
+        }
+        XCTAssertEqual(err?.localizedDescription, OptimizelyError.valueForKeyNotFound("").reason)
```
msohailhussain, line 256: is it the appropriate message? Question for Pawel/jae.

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
```diff
@@ -0,0 +1,489 @@
...
+            try self.optimizelyJSON.getValue(jsonPath: "field4.inner_field1", schema: &schema)
+        } catch {
+            err = error
+        }
+        XCTAssertEqual(err?.localizedDescription, OptimizelyError.failedToAssignValueToSchema.reason)
+    }
+    
+    func testGetValueForValidJSONMultipleKeyAndNonDecodableSchema() {
+        var err: Error?
+        var schema: [Any] = []
```
msohailhussain, line 317: Don't provide invalid schema, there msut be a valid schema and it should be non-decodable. may be a simple schema.

Sources/Data Model/OptimizelyJSON.swift
msohailhussain: yup, i believe so it should be public.

yasirfolio3: > Shouldn't we use `guard` if that's not a valid JSON object.

if it's not a valid json object, we then try to assign it directly to the schema, if both fail, we then throw `failedToAssignValueToSchema`.

Sources/Data Model/OptimizelyJSON.swift
yasirfolio3: Since we try to assign value directly to the generic operator if the above case fails, if both fail, then we throw a generalised error here:  
https://github.com/optimizely/swift-sdk/pull/315/files/b4bd90cd23dbc74c17edef1c776244e80ad55b16#diff-fbd9b0e0f2ff68ec3188bdee23588907R99

Sources/Data Model/OptimizelyJSON.swift
yasirfolio3: In case user provides schema such as `[Any]` which do not implement `Decodable` protocol, for such cases we have to overload the getValue method.

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
yasirfolio3: OptimizelyJSON should only accept `JSON Map` as parameter, `JSON Array` should not be acceptable.

Tests/OptimizelyTests-DataModel/OptimizelyJSONTests.swift
yasirfolio3: Since we explode jsonPath to array using `.` separator, `field4.` would result into an empty string on first index of array which will produce the above error with empty `key` value.

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -0,0 +1,143 @@
...
+                throw OptimizelyError.failedToConvertStringToDictionary
+            }
+            self.data = jsonData
+            self.payload = payload
+        } catch {
+            throw OptimizelyError.failedToConvertStringToDictionary
+        }
+    }
+    
+    @objc init(data: [String: Any]) throws {
```
jaeopt, line 41: We can return "nil" instead of throws. Same for other inits.

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -0,0 +1,143 @@
...
+* limitations under the License.                                           *
+***************************************************************************/
+
+import Foundation
+
+public class OptimizelyJSON: NSObject {
+    
+    private typealias schemaHandler = (Any) throws -> Void
+    private var payload: String?
+    private var data = [String: Any]()
```
jaeopt, line 23: What about "map" instead of "data"? data is too generic.

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -0,0 +1,143 @@
...
+        return data
+    }
+    
+    /// Populates the decodable schema passed by the user
+    ///
+    /// - Parameters:
+    ///   - jsonPath: Key path for the value.
+    ///   - schema: Decodable schema to populate.
+    /// - Throws: `OptimizelyError`
+    public func getValue<T: Decodable>(jsonPath: String, schema: UnsafeMutablePointer<T>) throws {
```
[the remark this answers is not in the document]
yasirfolio3: `inout` is not supported by objc neither are `generics`, this is a workaround to that.

Sources/Optimizely/OptimizelyJSON.swift
yasirfolio3: following `go` as reference PR, they have named it `data`, i feel `map` makes more sense too.

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -0,0 +1,143 @@
...
+            throw OptimizelyError.invalidDictionary
+        }
+        self.data = data
+    }
+    
+    // MARK: - OptimizelyJSON Implementation
+    
+    /// - Returns: The string representation of json
+    /// - Throws: `OptimizelyError`
+    @objc public func toString() throws -> String {
```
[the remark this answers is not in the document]
yasirfolio3: yes, agreed, il changed that.

Sources/Optimizely/OptimizelyJSON.swift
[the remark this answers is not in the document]
jaeopt: Can we use "inout" for swift core and use "UnsafePointer" for "ObjC" only. We can keep swift as clean as possible, and all dirty work for ObjC support in +ObjC codes.

Tests/OptimizelyTests-APIs/OptimizelyClientTests_OptimizelyJSON.swift
```diff
@@ -0,0 +1,510 @@
...
+    
+    func testGetValueForValidJSONKeyAndIncorrectDecodableStructSchema() {
+        var err: Error?
+        var schema = IncorrectSchema()
+        do {
+            try self.optimizelyJSON.getValue(jsonPath: "field4", schema: &schema)
+        } catch {
+            err = error
+        }
+        XCTAssertEqual(err?.localizedDescription, OptimizelyError.failedToAssignValueToSchema.reason)
```
pawels-optimizely: I don't think it should fail. if the user provides an empty or incorrect schema , it will not be parsed correctly but I don't think it should fail.

yasirfolio3: It fails since `IncorrectSchema` has a mandatory property defined as `var incorrectField: Int = 0`, On the other hand it will pass if it was defined as optional  `var incorrectField: Int?`.

yasirfolio3: We have defined `map` as `[String: Any]` which makes sure that values are non-nil hence i couldn't write test for `field6`, should we make them nullable? is there a case where a json string will have a `nil` value?

yasirfolio3: Adding a test for `field4..some_field` 👍

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -0,0 +1,142 @@
...
+        return getValue(jsonPath: jsonPath, schemaHandler: populateDecodableSchema(value:))
+    }
+    
+    /// Populates the schema passed by the user
+    ///
+    /// - Parameters:
+    ///   - jsonPath: Key path for the value.
+    ///   - schema: Schema to populate.
+    /// - Returns: true if value decoded successfully
+    public func getValue<T>(jsonPath: String, schema: UnsafeMutablePointer<T>) -> Bool {
```
jaeopt, line 105: Can we allow "nil" jsonPath same as ""? Nil is more convenient. We can assign default value to nil:
```suggestion
    public func getValue<T>(jsonPath: String? = nil, schema: UnsafeMutablePointer<T>) -> Bool {
```

yasirfolio3:
> LGTM

@jaeopt can you please merge this PR aswell?

PR #316, opened 2020-04-17
feat: added GetFeatureVariableJSON. (WIP)

yasirfolio3:
## Summary
- added GetFeatureVariableJSON API.

### DIFF
--- a/Sources/Data Model/FeatureVariable.swift
+++ b/Sources/Data Model/FeatureVariable.swift
@@ -1,5 +1,5 @@
 /****************************************************************************
-* Copyright 2019, Optimizely, Inc. and contributors                        *
+* Copyright 2019-2020, Optimizely, Inc. and contributors                   *
 *                                                                          *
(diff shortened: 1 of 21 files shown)


Sources/Data Model/FeatureVariable.swift
```diff
@@ -20,6 +20,36 @@ struct FeatureVariable: Codable, Equatable {
     var id: String
     var key: String
     var type: String
+    var subType: String?
     // datafile schema rquires this, but test has "null" value case. keep optional for FSC
     var defaultValue: String?
+    
+    enum CodingKeys: String, CodingKey {
```
msohailhussain, line 27: why we need this Encoding stuff?

yasirfolio3: to manually assign value to `type`:
```
if type == "string" && subType == "json" {
     type = "json"
 }
```

PR #317, opened 2020-04-22
feat: added GetFeatureVariableJSON.

yasirfolio3:
## Summary
- added GetFeatureVariableJSON API.

## Test Plan
- FSC should pass.
- Added Unit tests.

### DIFF
--- a/Sources/Data Model/FeatureVariable.swift
+++ b/Sources/Data Model/FeatureVariable.swift
@@ -1,5 +1,5 @@
 /****************************************************************************
-* Copyright 2019, Optimizely, Inc. and contributors                        *
+* Copyright 2019-2020, Optimizely, Inc. and contributors                   *
 *                                                                          *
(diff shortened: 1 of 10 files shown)


Sources/Data Model/FeatureVariable.swift
```diff
@@ -20,6 +20,36 @@ struct FeatureVariable: Codable, Equatable {
...
     var defaultValue: String?
+    
+    enum CodingKeys: String, CodingKey {
+        case id
+        case key
+        case type
+        case subType
+        case defaultValue
+    }
+
```
jaeopt, line 34: We can drop this since all default names

yasirfolio3: CodingKeys (Even if exactly the same as defaultNames) are required to be defined if we override `init(from decoder: Decoder)`.

PR #319, opened 2020-04-27
refact(OptimizelyJSON): Refactored to return decoded value.

yasirfolio3:
## Summary
- Refactored `getValue` to return decoded value instead of setting it to schema passed as parameter.

## Test Plan
- FSC should pass.
- Updated Unit tests.

### DIFF
--- a/Sources/Optimizely/OptimizelyJSON+ObjC.swift
+++ b/Sources/Optimizely/OptimizelyJSON+ObjC.swift
@@ -1,0 +1,58 @@
+/****************************************************************************
+* Copyright 2020, Optimizely, Inc. and contributors                        *
+*                                                                          *
+* Licensed under the Apache License, Version 2.0 (the "License");          *
+* you may not use this file except in compliance with the License.         *
+* You may obtain a copy of the License at                                  *
+*                                                                          *
+*    http://www.apache.org/licenses/LICENSE-2.0                            *
+*                                                                          *
+* Unless required by applicable law or agreed to in writing, software      *
+* distributed under the License is distributed on an "AS IS" BASIS,        *
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+
+import Foundation
+
+extension OptimizelyJSON {
+    
+    @available(swift, obsoleted: 1.0)
+    @objc(initWithPayload:)
+    convenience init?(p: String) {
+        self.init(payload: p)
+    }
+    
+    @available(swift, obsoleted: 1.0)
+    @objc(initWithMap:)
+    convenience init?(m: [String: Any]) {
+        self.init(map: m)
+    }
+    
+    @available(swift, obsoleted: 1.0)
+    @objc(toString)
+    /// - Returns: The string representation of json
+    public func objcToString() -> String? {
+        return self.toString()
+    }
+    
+    @available(swift, obsoleted: 1.0)
+    @objc(toMap)
+    /// - Returns: The json dictionary
+    public func objcToMap() -> [String: Any] {
+        return self.toMap()
+    }
+    
+    @available(swift, obsoleted: 1.0)
+    @objc(getValueWithJsonPath:schema:)
+    /// Populates the schema passed by the user
+    ///
+    /// - Parameters:
+    ///   - jsonPath: Key path for the value.
+    ///   - schema: Schema to populate.
+    /// - Returns: true if value decoded successfully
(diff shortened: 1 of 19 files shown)


Tests/OptimizelyTests-APIs/OptimizelyClientTests_OptimizelyJSON_Objc.m
```diff
@@ -63,13 +63,13 @@ - (void)testGetValue {
...
-    XCTAssertTrue([optimizelyJSON getValueWithJsonPath:@"testfield" schema: &actualIntValue]);
+    actualIntValue = [optimizelyJSON getValueWithJsonPath:@"testfield"];
     XCTAssertEqual(expectedValue, actualIntValue);
     
     // Fetching dictionary type
     optimizelyJSON = [[OptimizelyJSON alloc] initWithMap:_map];
     id actualDictionaryValue = @{};
-    XCTAssertTrue([optimizelyJSON getValueWithJsonPath:@"" schema: &actualDictionaryValue]);
+    actualDictionaryValue = [optimizelyJSON getValueWithJsonPath:@""];
     XCTAssertTrue([_map isEqualToDictionary:(NSDictionary *)actualDictionaryValue]);
```
jaeopt, line 73: Can we add more test cases? Are we allowing "Dictionary" type only? What about user-defined class type?

yasirfolio3: Do we need to convert all swift tests for objc? even though objc is internally calling the swift method? We have covered all cases for swift.

jaeopt:
I do not think so. We just need the public APIs work fine.
We already covered reading primitive data types and map.

The test cases we need is:
ObjC custom class (instead of NSDictionary) work with
getFeatureVariabaleJSON.

*Jae Kim*
Sr. Software Engineer
e. jae.kim@optimizely.com

Tests/OptimizelyTests-APIs/OptimizelyClientTests_OptimizelyJSON_Objc.m
jaeopt: I do not think so. We just need show the public APIs work fine. 
We already covered reading primitive data types and map.
 
The test cases we need is:
ObjC custom class (instead of NSDictionary) work with getFeatureVariabaleJSON.

yasirfolio3: ObjectiveC does not support Codable protocol, user will have to get value into an `NSDictionary` and parse it himself which is why i didn't write test related to custom class. I tried writing a test with a custom objc class which implemented NSCodable but swift wasn't able to assigb values into that aswell.

Sources/Optimizely/OptimizelyJSON+ObjC.swift
```diff
@@ -45,14 +45,13 @@ extension OptimizelyJSON {
...
     ///
     /// - Parameters:
     ///   - jsonPath: Key path for the value.
-    ///   - schema: Schema to populate.
-    /// - Returns: true if value decoded successfully
-    public func objcGetValue(jsonPath: String, schema: UnsafeMutablePointer<AnyObject>) -> Bool {
-        return self.getValue(jsonPath: jsonPath, schema: &schema.pointee)
+    /// - Returns: Value if decoded successfully
+    public func objcGetValue(jsonPath: String) -> Any? {
+        return self.getValue(jsonPath: jsonPath)
```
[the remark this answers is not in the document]
yasirfolio3: If we kept the previous implementation then we will have to make `schema` as optional which will force user to declare schema as nullable. Then 
`schema.pointee =  self.getValue(jsonPath: jsonPath)` should work.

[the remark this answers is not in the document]
jaeopt: If custom type cannot be supported, getValue() does not provide much help, just adding confusion. Let's remove it from ObjC interface. So ObjC users will see "toString()" and "toMap()" interface only. They can use "toMap()" call and walk through to get the target value.

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -100,9 +100,9 @@ public class OptimizelyJSON: NSObject {
     
     /// Returns parsed value for jsonPath
     ///
-    /// If JSON Data is {"k1":true, "k2":{"k3":"v2"}}
+    /// If JSON Data is {"k1":true, "k2":{"k3":"v3"}}
     ///
-    /// Set jsonPath to "k1" to access the true boolean value or set it to to "k2.k3" to access {"k3":"v2"}.
+    /// Set jsonPath to "k2" to access {"k3":"v3"} or set it to to "k2.k3" to access "v3"
```
jaeopt, line 105: Fix "to to"

PR #320, opened 2020-05-05
Forward resourceTimeout to datafileHandler

ffittschen:
During evaluations of which `start` method we should use in our project, we discovered that in the async initialization, the `resourceTimeout` is not forwarded to the `datafileHandler`. This will be fixed with this PR.


jaeopt:
@ffittschen It' a good catch! Can you sign our CLA if you did not yet?
https://github.com/optimizely/swift-sdk/blob/master/CONTRIBUTING.md

mikeng13:
This actually ends up breaking tests:

image

Could you please fix the compile error in there?

ffittschen:
Build should be green now. There are still some tests failing locally, but those same tests also fail on `master`.
I also signed the CLA

PR #323, opened 2020-05-12
fix: Targeting rollout should move to fallback rule when bucketing fails.

yasirfolio3:
## Summary
- fixes target-rollout error continuing other rules when a rule hits but bucketing fails.

## Test Plan
- FSC should pass.
- Updated unit tests accordingly.

## Issues
- https://optimizely.atlassian.net/browse/OASIS-6334

### DIFF
--- a/Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
+++ b/Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
@@ -1,5 +1,5 @@
 /****************************************************************************
-* Copyright 2019, Optimizely, Inc. and contributors                        *
+* Copyright 2019-2020, Optimizely, Inc. and contributors                   *
 *                                                                          *
 * Licensed under the Apache License, Version 2.0 (the "License");          *
 * you may not use this file except in compliance with the License.         *
--- a/Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
+++ b/Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
@@ -39,11 +39,21 @@ class DecisionServiceTests_Features: XCTestCase {
     var kAudienceIdAge = "20"
     var kAudienceIdInvalid = "9999999"
     
+    var kRolloutVariationKeyA = "rolloutA"
+    var kRolloutVariationKeyB = "rolloutB"
+    var kRolloutVariationKeyC = "rolloutC"
+    
+    var kRolloutAudienceIdAge1 = "30"
+    var kRolloutAudienceIdAge2 = "40"
+    
     var kAttributesCountryMatch: [String: Any] = ["country": "us"]
     var kAttributesCountryNotMatch: [String: Any] = ["country": "ca"]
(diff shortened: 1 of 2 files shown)


Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
```diff
@@ -127,32 +129,78 @@ class DecisionServiceTests_Features: XCTestCase {
...
+                        "entityId": "10389700000",
+                        "endOfRange": 10000
+                    ]
+                ],
+                "audienceIds": [kAudienceIdAge],
+                "variations": [
+                    [
+                        "variables": [],
+                        "id": "10389700000",
+                        "key": kVariationKeyA,
```
jaeopt, line 154: Let's use different key names (like "kRolloutVariationKeyA"). It's already used in experiment variations. Can be confused. Same for other rollout variations as well.

Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
```diff
@@ -253,10 +301,58 @@ extension DecisionServiceTests_Features {
...
     
+    func testGetVariationForFeatureRolloutEvaluatesNextIfAudienceEvaluationFails() {
+        self.config.project.rollouts = [try! OTUtils.model(from: sampleRolloutData)]
+        featureFlag.rolloutId = kRolloutId
+        self.config.project.featureFlags = [featureFlag]
+        
+        let variation = self.decisionService.getVariationForFeatureRollout(config: config,
+                                                                           featureFlag: featureFlag,
+                                                                           userId: kUserId,
+                                                                           attributes: kAttributesCountryMatch)
```
jaeopt, line 325: I think, we need different audience for 2nd rollout rule. CountryMatch is used for feature-experiment. So no guarantee if it's passed down to rollout.

Tests/OptimizelyTests-Common/DecisionServiceTests_Features.swift
```diff
@@ -253,10 +301,58 @@ extension DecisionServiceTests_Features {
...
+    func testGetVariationForFeatureRolloutFallbackRule() {
+        self.config.project.rollouts = [try! OTUtils.model(from: sampleRolloutData)]
+        self.config.project.rollouts[0].experiments[0].trafficAllocation[0].endOfRange = 0
+        featureFlag.rolloutId = kRolloutId
+        self.config.project.featureFlags = [featureFlag]
+        
+        let variation = self.decisionService.getVariationForFeatureRollout(config: config,
+                                                                           featureFlag: featureFlag,
+                                                                           userId: kUserId,
+                                                                           attributes: kAttributesAgeMatch)
```
jaeopt, line 313: With this setting, we cannot tell if it skips the rest of them or goes thru all rules. What about this setting:

Rule1: age < 30 (traffic allocation = 0)
Rule2: age < 40 (traffic allocation = 10000)
Rule3: everybody

When attribute age = 20, we should expect variation for Rule 3 if it skips. Otherwise, variation for Rule 2 will hit.

PR #325, opened 2020-05-12
chore: turn on build-libraries-for-distribution for ABI support

jaeopt:
Turn on "Build Libraries For Distribution" for ABI support
* SDK users can build a binary and share it. 
* ABI should be enabled for supporting it even if the sdk is source distributed.


Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 159491f53ea704c30be049e048d854a49f9b7433
 
-COCOAPODS: 1.9.1
```
thomaszurkan-optimizely, line 16: did you have to downgrade cocoapods or did you not want to install the latest?

PR #333, opened 2020-06-18
feature: add semantic version types and test

thomaszurkan-optimizely:
## Summary
- add greater than or equal and less than or equal for numbers
- add semantic version match types.
## Teset
Add unit tests


aliabbasrizvi:
Not sure if we have introduced compatibility suite tests for this. Can you confirm? I think may be we can add that first before going down this path?

Sources/Data Model/Audience/SemanticVersion.swift
```diff
@@ -0,0 +1,62 @@
...
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+
+import Foundation
+
+typealias SemanticVersion = String
+/*
+ This comparison is ported from the Optimizely web version of semantic version compare.
+ https://github.com/optimizely/client-js/blob/devel/src/core/lib/compare_version.js
```
aliabbasrizvi, line 22: This is a private repo. We should ideally drop the link here.

aliabbasrizvi: We should probably add a note that beta/alpha/any other string which does not resemble x.y.z will not be evaluated here.

Tests/OptimizelyTests-DataModel/AttributeValueTests_Evaluate.swift
```diff
@@ -256,5 +256,38 @@ extension AttributeValueTests_Evaluate {
     func allValueTypes(value: Double) -> [Any] {
         return allIntTypes(value: value) + allNumTypes(value: value)
     }
+ }
+
+// MARK: - SemanticVersion
+extension AttributeValueTests_Evaluate {
+    func testIsSemanticSame() {
+        let model = try! OTUtils.getAttributeValueFromNative("2.0.0")
+        XCTAssertTrue(try! model.isSemanticVersionEqual(than: "2.0"))
```
aliabbasrizvi, line 265: nit. Can you add a note here just to distinguish from the next test.

Tests/OptimizelyTests-DataModel/UserAttributeTests_Evaluate.swift
```diff
@@ -264,5 +289,57 @@ extension UserAttributeTests_Evaluate {
...
+        XCTAssertTrue(try! model.evaluate(attributes: attributes))
+        attributes["version"] = "2.9"
+        XCTAssertTrue(try! model.evaluate(attributes: attributes))
+        attributes["version"] = "1.9"
+        XCTAssertFalse(try! model.evaluate(attributes: attributes))
+    }
 
 }
+
+
```
aliabbasrizvi, line 345: nit. remove extraneous new lines.

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+
+class SemanticVersionTests: XCTestCase {
+    
+    // MARK: - version tests.
+    
+    func testTargetString() {
+        let target = "2.0" as SemanticVersion
+        let version = "2.0.1" as SemanticVersion
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) == 0)
```
jaeopt, line 29: Is this correct? I expect 2.0 < 2.0.1

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+        let version = "2.9.1" as SemanticVersion
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) > 0)
+    }
+
+    func testTargetMinorPartEqual() {
+        let target = "2.9" as SemanticVersion
+        let version = "2.9.1" as SemanticVersion
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) == 0)
```
jaeopt, line 84: same here. I expect 2.9.1 > 2.9 = 2.9.0

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    func testTargetPatchWithBetaTagEqual() {
+        let target = "2.9.9-beta" as SemanticVersion
+        let version = "2.9.9-beta" as SemanticVersion
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) == 0)
+    }
+
+    func testPartialVersionEqual() {
```
jaeopt, line 115: Can we add tests:
- 1.0.0 = 1.0.0-beta
- 1.0.0 < 1.1.0-beta
- 1.2.0-beta > 1.1.0
- 1.2.0-beta < 1.2.1

jaeopt: Also need to test with different "beta" rules: 1.2.0.beta, 1.2.0-beta

Sources/Data Model/Audience/UserAttribute.swift
```diff
@@ -129,10 +136,44 @@ extension UserAttribute {
...
+        case .semver_gt:
+            guard let targetValue = AttributeValue(value: rawAttributeValue) else {
+                throw OptimizelyError.evaluateAttributeInvalidType("attribute value \(nameFinal) invalid type")
+             }
+            return try targetValue.isSemanticVersionGreater(than: value!.stringValue)
+        case .semver_ge:
+            guard let targetValue = AttributeValue(value: rawAttributeValue) else {
+                throw OptimizelyError.evaluateAttributeInvalidType("attribute value \(nameFinal) invalid type")
+             }
+            return try targetValue.isSemanticVersionGreaterOrEqual(than: value!.stringValue)
```
jaeopt, line 176: We can change them (reverse compare) to be consistent with others?

```suggestion
        case .semver_eq:
            return try value!.isSemanticVersionEqual(than: rawAttributeValue!)
        case .semver_lt:
            return try value!.isSemanticVersionGreater(than: rawAttributeValue!)
        case .semver_le:
            return try value!.isSemanticVersionGreaterOrEqual(than: rawAttributeValue!)
        case .semver_gt:
            return try value!.isSemanticVersionLess(than: rawAttributeValue!)
        case .semver_ge:
            return try value!.isSemanticVersionLessOrEqual(than: rawAttributeValue!)
```

Sources/Data Model/Audience/SemanticVersion.swift
```diff
@@ -0,0 +1,62 @@
...
+          }
+
+          // Expect a version string of the form x.y.z
+        let targetedVersionParts = targetedVersion.split(separator: ".");
+        let versionParts = self.split(separator: ".");
+
+        // Up to the precision of targetedVersion, expect version to match exactly.
+        for (idx, _) in targetedVersionParts.enumerated() {
+            if versionParts.count <= idx {
+              return -1;
```
jaeopt, line 39: This is different from what I expected. Can you double check the version compare rules?
I expect:
- 1.2 = 1.2.0 > 1.2.1
- 1.2 > 1.2.1-beta = 1.2.1.beta

Sources/Data Model/Audience/SemanticVersion.swift
thomaszurkan-optimizely: @aliabbasrizvi take a closer look at the implementation.  it does allow for string tags.

Sources/Data Model/Audience/UserAttribute.swift
thomaszurkan-optimizely: i thought about this but i think it adds too much confusion on the semantic version.  i believe this is a better logical split even if it deviates from the others slightly.

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
thomaszurkan-optimizely: depends on which is target

Tests/OptimizelyTests-DataModel/AttributeValueTests_Evaluate.swift
```diff
@@ -256,5 +256,38 @@ extension AttributeValueTests_Evaluate {
...
+    }
+
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: @jaeopt  it depends on which one is the attribute and which is the match value/target.  The target is always passed into the compare.

Sources/Data Model/Audience/UserAttribute.swift
```diff
@@ -142,43 +142,43 @@ extension UserAttribute {
...
         case .ge:
             // user attribute "greater than or equal" this condition value
             // so evaluate if this condition value "isLess" than or equal the user attribute value
-            return try value!.isLessOrEqual(than: rawAttributeValue!)
+            return try value!.isLessOrEqual(than: rawAttributeValue!, condition: stringRepresentation, name: nameFinal)
         // semantic versioning seems unique.  the comarison is to compare verion but the passed in version is the target version.
         case .semver_eq:
             guard let targetValue = AttributeValue(value: rawAttributeValue) else {
-                throw OptimizelyError.evaluateAttributeInvalidType("attribute value \(nameFinal) invalid type")
+                throw OptimizelyError.evaluateAttributeInvalidCondition("attribute value \(nameFinal) invalid type")
```
jaeopt, line 161: Don't we need sem-version type checking and log error cases? "invalid sem-ver format", etc?

Tests/OptimizelyTests-DataModel/AttributeValueTests_Evaluate.swift
```diff
@@ -260,34 +260,49 @@ extension AttributeValueTests_Evaluate {
...
         let model = try! OTUtils.getAttributeValueFromNative("2.1.9")
         XCTAssertTrue(try! model.isSemanticVersionGreater(than: "2.1.6"))
     }
+    
+    // Test compare equal when target is major.minor.patch-beta
+    func testIsSemanticFullEqual() {
+        let model = try! OTUtils.getAttributeValueFromNative("2.1.9-beta")
+        XCTAssertTrue(try! model.isSemanticVersionEqual(than: "2.1.9-beta"))
+    }
+
```
jaeopt, line 307: Can we add more beta tests?
- 2.1.3-beta > target(2.1.2)
- 2.1.3-beta = target(2.1.3) [I think this is correct since we assume they have same feature]
- 2.1.3-beta < target(2.1.4)

same tests fro 2.1.3.beta

jaeopt: I also like to see tests (if I understand sem-ver rule correctly).
- 2.1.0 = target(2.1)
- 2.1.215 = target(2.1)
- 2.12 = target(2)
- 2.785.13 = target(2)

Sources/Data Model/Audience/SemanticVersion.swift
```diff
@@ -0,0 +1,61 @@
...
+            } else if !String(versionParts[idx]).isNumber {
+                //Compare strings
+                if versionParts[idx] != targetedVersionParts[idx] {
+                    return -1;
```
jaeopt, line 42: Is this correct?  I expect: 3.1.2-beta > 3.1.0, but this returns opposite.
When it's required for a client to turn on a feature for version > 3.1.0, this rule should be true to all beta version after the target as well.

Sources/Data Model/Audience/SemanticVersion.swift
```diff
@@ -0,0 +1,61 @@
...
+            return 0
+          }
+
+          // Expect a version string of the form x.y.z
+        let targetedVersionParts = targetedVersion.split(separator: ".");
+        let versionParts = self.split(separator: ".");
+
+        // Up to the precision of targetedVersion, expect version to match exactly.
+        for (idx, _) in targetedVersionParts.enumerated() {
+            if versionParts.count <= idx {
```
msohailhussain, line 37: What we want to achieve using this condition? Based on count and compare with idx looks weird

Sources/Data Model/Audience/AttributeValue.swift
```diff
@@ -182,6 +186,32 @@ extension AttributeValue {
...
+    func isSemanticVersionGreater(than target: SemanticVersion) throws -> Bool {
+        return (self.stringValue as SemanticVersion).compareVersion(targetedVersion: target) > 0
+    }
+
+    func isSemanticVersionLess(than target: SemanticVersion) throws -> Bool {
+        return (self.stringValue as SemanticVersion).compareVersion(targetedVersion: target) < 0
+    }
+
+    func isSemanticVersionGreaterOrEqual(than target: SemanticVersion) throws -> Bool {
+        return try isSemanticVersionGreater(than: target) ||
```
msohailhussain, line 206: @thomaszurkan-optimizely What about 
`(self.stringValue as SemanticVersion).compareVersion(targetedVersion: target)` >= 0
and same for less than <=
we can save one call when version is equal.

Sources/Data Model/Audience/SemanticVersion.swift
msohailhussain: yeah I need to know the same thing. 
What about comparing `alpha` and `beta`

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
thomaszurkan-optimizely: 1.2.0.beta is not valid semver

Tests/OptimizelyTests-DataModel/SemanticVersionTests.swift
```diff
@@ -118,4 +118,41 @@ class SemanticVersionTests: XCTestCase {
...
+        let version = "2.1.3-beta"
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) > 0)
+    }
+    
+    func testBetaToRelease() {
+        let target = "2.1.2-release"
+        let version = "2.1.2-beta"
+        
+        XCTAssert(version.compareVersion(targetedVersion: target) < 0)
```
jaeopt, line 133: Is this test valid? I thought "-release" cannot be in the target?

thomaszurkan-optimizely: both of those are valid suffixes in the semver pre-release www.semver.org

PR #334, opened 2020-06-18
refact(audience-logs): Added and refactored audience evaluation logs.

yasirfolio3:
## Summary
- Added and updated missing audience evaluation logs.

## Issues
https://optimizely.atlassian.net/browse/OASIS-6545

### DIFF
--- a/Sources/Data Model/Audience/Audience.swift
+++ b/Sources/Data Model/Audience/Audience.swift
@@ -1,5 +1,5 @@
 /****************************************************************************
-* Copyright 2019, Optimizely, Inc. and contributors                        *
+* Copyright 2019-2020, Optimizely, Inc. and contributors                   *
 *                                                                          *
 * Licensed under the Apache License, Version 2.0 (the "License");          *
 * you may not use this file except in compliance with the License.         *
--- a/Sources/Data Model/Audience/Audience.swift
+++ b/Sources/Data Model/Audience/Audience.swift
@@ -64,6 +64,21 @@ struct Audience: Codable, Equatable {
     }
     
     func evaluate(project: ProjectProtocol?, attributes: OptimizelyAttributes?) throws -> Bool {
-        return try conditions.evaluate(project: project, attributes: attributes)
+        let logger = OPTLoggerFactory.getLogger()
+        if let conditions = try? JSONEncoder().encode(conditions), let conditionsString = String(data: conditions, encoding: .utf8)  {
+            logger.d(.audienceEvaluationStarted(id, conditionsString))
+        }
+        var result: Bool?
+        var err: Error?
+        do {
+            result = try conditions.evaluate(project: project, attributes: attributes)
+        } catch {
+            err = error
+        }
+        logger.d(.audienceEvaluationResult(id, result?.description ?? "unknown"))
+        guard let error = err else {
+            return result!
+        }
+        throw error
     }
 }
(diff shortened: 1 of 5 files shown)


Sources/Data Model/Audience/Audience.swift
```diff
@@ -64,6 +64,21 @@ struct Audience: Codable, Equatable {
     }
     
     func evaluate(project: ProjectProtocol?, attributes: OptimizelyAttributes?) throws -> Bool {
-        return try conditions.evaluate(project: project, attributes: attributes)
+        let logger = OPTLoggerFactory.getLogger()
```
jaeopt, line 67: A separate logger instance in each Audience must be a waste.
Let's create a single instance in Project and pass it as a parameter here:

func evaluate(project: ProjectProtocol?, attributes: OptimizelyAttributes?, logger: OPTLogger? = nil)

Sources/Data Model/Audience/Audience.swift
```diff
@@ -64,6 +64,21 @@ struct Audience: Codable, Equatable {
...
+            result = try conditions.evaluate(project: project, attributes: attributes)
+        } catch {
+            err = error
+        }
+        logger.d(.audienceEvaluationResult(id, result?.description ?? "unknown"))
+        guard let error = err else {
+            return result!
+        }
+        throw error
     }
```
jaeopt, line 83:
```suggestion
      let result = try conditions.evaluate(project: project, attributes: attributes)
      logger?.d(.audienceEvaluationResult(id, result.description))
      return result
```

Sources/Data Model/Audience/Audience.swift
```diff
@@ -64,6 +64,21 @@ struct Audience: Codable, Equatable {
...
     
     func evaluate(project: ProjectProtocol?, attributes: OptimizelyAttributes?) throws -> Bool {
-        return try conditions.evaluate(project: project, attributes: attributes)
+        let logger = OPTLoggerFactory.getLogger()
+        if let conditions = try? JSONEncoder().encode(conditions), let conditionsString = String(data: conditions, encoding: .utf8)  {
```
jaeopt, line 68: Let's change this for efficiency (skipping whole parsing and string prepare for lower log-level)
```suggestion
        logger?.d {
            if let conditions = try? JSONEncoder().encode(conditions), let conditionsString = String(data: conditions, encoding: .utf8) {
                return LogMessage.audienceEvaluationStarted(id, conditionsString).description
            } else {
                return "Invalid conditions format"
            }
        }
```

Sources/Data Model/Audience/Audience.swift
jaeopt: Instantiating "OPTLogger" in "Project" will make "Project" less clean, but we can use the same logger for other logging later.

Sources/Data Model/Audience/Audience.swift
```diff
@@ -64,6 +64,21 @@ struct Audience: Codable, Equatable {
...
+            logger.d(.audienceEvaluationStarted(id, conditionsString))
+        }
```
jaeopt, line 70: Let's change this for efficiency (skipping whole parsing and string prepare for lower log-level)

```suggestion
        logger?.d {
            if let conditions = try? JSONEncoder().encode(conditions), let conditionsString = String(data: conditions, encoding: .utf8) {
                return LogMessage.audienceEvaluationStarted(id, conditionsString).description
            } else {
                return "Invalid conditions format"
            }
        }
```

PR #336, opened 2020-06-29
refact(audience-logs): Added and refactored audience evaluation logs.

yasirfolio3:
## Summary
- Added and updated missing audience evaluation logs.

## Issues
https://optimizely.atlassian.net/browse/OASIS-6545


msohailhussain:
I have done my first pass, please address.

Sources/Data Model/Project.swift
```diff
@@ -42,6 +42,23 @@ struct Project: Codable, Equatable {
     var typedAudiences: [Audience]?
     var featureFlags: [FeatureFlag]
     var botFiltering: Bool?
+    let logger = OPTLoggerFactory.getLogger()
+    
+    // Required since logger in not decodable
+    enum CodingKeys: String, CodingKey {
```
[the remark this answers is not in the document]
yasirfolio3: Adding `let logger = OPTLoggerFactory.getLogger()` caused the struct to throw error since logger does not confirm to `Codable` and `Equatable` protocols. The solution is to tell the struct explicitly which properties will confirm to `Codable` and `Equatable`.

Sources/Data Model/Audience/AttributeValue.swift
```diff
@@ -231,7 +256,7 @@ extension AttributeValue {
 
         // valid range: [-2^53, 2^53]
         if abs(num) > pow(2, 53) {
-            throw OptimizelyError.evaluateAttributeValueOutOfRange(prettySrc(caller, target: number))
+            throw OptimizelyError.evaluateAttributeValueOutOfRange(condition, name)
```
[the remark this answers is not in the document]
yasirfolio3: Updated it as per python since it also doesn't log the source of the error.

Sources/Data Model/Project.swift
```diff
@@ -42,6 +42,23 @@ struct Project: Codable, Equatable {
     var typedAudiences: [Audience]?
     var featureFlags: [FeatureFlag]
     var botFiltering: Bool?
+    let logger = OPTLoggerFactory.getLogger()
```
jaeopt, line 45: Can you add a space above to make it clear from the other JSON fields?

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -18,6 +18,11 @@ import Foundation
 
 class DefaultDecisionService: OPTDecisionService {
     
+    enum evaluationLogType: String {
```
jaeopt, line 21:
```suggestion
    enum EvaluationLogType: String {
```

Sources/Utils/Utils.swift
```diff
@@ -128,7 +126,7 @@ class Utils {
     }
     
     static func getConditionString<T: Encodable>(conditions: T) -> String {
-        if let jsonData = try? self.jsonEncoder.encode(conditions), let jsonString = String(data: jsonData, encoding: .utf8) {
+        if let jsonData = try? JSONEncoder().encode(conditions), let jsonString = String(data: jsonData, encoding: .utf8) {
```
msohailhussain, line 129: This is expensive to initialize JSONEncoder on everycall, I think you may keep `static` inside the func. @jaeopt do you have any suggestion?

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -80,12 +85,19 @@ class DefaultDecisionService: OPTDecisionService {
...
-    func isInExperiment(config: ProjectConfig, experiment: Experiment, userId: String, attributes: OptimizelyAttributes) -> Bool {
+    func isInExperiment(config: ProjectConfig, experiment: Experiment, userId: String, attributes: OptimizelyAttributes, logType: evaluationLogType = .experiment, loggingKey: String? = nil) -> Bool {
         
         var result = true   // success as default (no condition, etc)
+        let evType = logType.rawValue
+        
+        var finalLoggingKey = experiment.key
+        if let key = loggingKey {
+            finalLoggingKey = key
+        }
```
jaeopt, line 96:
```suggestion
        let finalLoggingKey = loggingKey ?? experiment.key
```

Sources/Implementation/DefaultDecisionService.swift
msohailhussain, line 21: Don't we have separate file for enums?

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -80,12 +85,19 @@ class DefaultDecisionService: OPTDecisionService {
...
         
         do {
             if let conditions = experiment.audienceConditions {
+                logger.d(.evaluatingAudiencesCombined(evType, finalLoggingKey, Utils.getConditionString(conditions: conditions)))
```
jaeopt, line 100:
```suggestion
                logger.d {
                   return .evaluatingAudiencesCombined(evType, finalLoggingKey, Utils.getConditionString(conditions: conditions)))
                }
```

"getConditionsString" will be expensive. Let's avoid this for normal flow (logLevel < .debug).
https://github.com/optimizely/swift-sdk/blob/bc787345a2932833d88ac9881ec6fa2f3bcdc231/Sources/Customization/Protocols/OPTLogger.swift#L51

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -100,23 +112,23 @@ class DefaultDecisionService: OPTDecisionService {
                     result = true
                 }
             }
-            // backward compatibility with audiencIds list
+                // backward compatibility with audiencIds list
```
jaeopt, line 115: nit: fix aligment

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -100,23 +112,23 @@ class DefaultDecisionService: OPTDecisionService {
...
             else if experiment.audienceIds.count > 0 {
                 var holder = [ConditionHolder]()
                 holder.append(.logicalOp(.or))
                 for id in experiment.audienceIds {
                     holder.append(.leaf(.audienceId(id)))
                 }
-                
+                logger.d(.evaluatingAudiencesCombined(evType, finalLoggingKey, Utils.getConditionString(conditions: holder)))
```
jaeopt, line 122: Fix to logger.d{} for all non-trivial messages.

Sources/Data Model/Audience/UserAttribute.swift
```diff
@@ -90,49 +90,55 @@ extension UserAttribute {
     
     func evaluate(attributes: OptimizelyAttributes?) throws -> Bool {
         
+        let conditionString = Utils.getConditionString(conditions: self)
```
jaeopt, line 93: This conditionString buidling is expensive. Let's remove this from normal flow (non-error).
We can build it in each error case.

Sources/Optimizely/OptimizelyError.swift
```diff
@@ -48,12 +48,16 @@ public enum OptimizelyError: Error {
...
+    case evaluateAttributeInvalidType(_ condition: String, _ value: Any, _ key: String)
+    case evaluateAttributeValueOutOfRange(_ condition: String, _ key: String)
     case evaluateAttributeInvalidFormat(_ hint: String)
-    case userAttributeInvalidType(_ hint: String)
-    case userAttributeInvalidMatch(_ hint: String)
-    case userAttributeInvalidFormat(_ hint: String)
+    case userAttributeInvalidType(_ condition: String)
+    case userAttributeInvalidMatch(_ condition: String)
+    case userAttributeNilValue(_ condition: String)
+    case nilAttributeValue(_ condition: String, _ key: String)
```
jaeopt, line 58: Do we need both of these? Looks like redundant. If so, let's use "userAttributeNilValue" to be consistent.

Sources/Data Model/Audience/UserAttribute.swift
```diff
@@ -90,49 +90,55 @@ extension UserAttribute {
...
+                throw OptimizelyError.nilAttributeValue(conditionString, nameFinal)
             }
         }
         
         switch matchFinal {
         case .exists:
             return !(rawAttributeValue is NSNull || rawAttributeValue == nil)
         case .exact:
-            return try value!.isExactMatch(with: rawAttributeValue!)
+            return try value!.isExactMatch(with: rawAttributeValue!, condition: conditionString, name: nameFinal)
```
jaeopt, line 131: This case is tough. How can we avoid "conditionString" buiding overhead?

Sources/Utils/Utils.swift
yasirfolio3: declaring static property inside a method throws this error: `Static properties may only be declared on a type`, solution to this is declaring `private static let jsonEncoder = JSONEncoder()` inside the class like i had implemented earlier.

yasirfolio3: declaring static property inside a method throws this error: `Static properties may only be declared on a type`, Workaround to this is declaring the property inside the class, similar to how it was done in the first commit.

Sources/Optimizely/OptimizelyError.swift
yasirfolio3: both serve different purposes: 
`userAttributeNilValue` = `Audience condition (\(condition)) evaluated to UNKNOWN because of null value.`
`nilAttributeValue` = `Audience condition (\(condition)) evaluated to UNKNOWN because a null value was passed for user attribute (\(key)).`

Sources/Data Model/Audience/UserAttribute.swift
yasirfolio3: Tried several ideas to evaluate `conditionString` just once for each `UserAttribute`. The problem i faced is that we are dealing with a struct which is a value type and to have a lazy or mutable property inside it, we would have to mark `evaluate` method as `mutating` which is not possible in our case since it is part of a protocol.

Sources/Optimizely/OptimizelyError.swift
```diff
@@ -118,13 +122,16 @@ extension OptimizelyError: CustomStringConvertible {
...
-        case .userAttributeInvalidMatch(let hint):          message = "UserAttribute has an invalid match type (\(hint))"
-        case .userAttributeInvalidFormat(let hint):         message = "UserAttribute has an invalid format (\(hint))"
-
+        case .userAttributeInvalidType(let condition):      message = "Audience condition (\(condition)) uses an unknown condition type. You may need to upgrade to a newer release of the Optimizely SDK."
+        case .userAttributeInvalidMatch(let condition):     message = "Audience condition (\(condition)) uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK."
+        case .userAttributeNilValue(let condition):         message = "Audience condition (\(condition)) evaluated to UNKNOWN because of null value."
+        case .nilAttributeValue(let condition, let key):    message = "Audience condition (\(condition)) evaluated to UNKNOWN because a null value was passed for user attribute (\(key))."
+        case .userAttributeInvalidName(let condition):      message = "Audience condition (\(condition)) evaluated to UNKNOWN because of invalid attribute name."
```
jaeopt, line 133: I see we need both. Can we change the order of these two lines so "userAttribute" and "attribute" errors can be grouped?

Sources/Data Model/Audience/UserAttribute.swift
jaeopt: One possible fix is to create "conditionsString" when UserAttribute is instantiated by datafile parsing (just like we did for "FeatureVariable.swift" for JSON type conversion. Then we can avoid "Utils.getConditionString".

Tests/OptimizelyTests-DataModel/AttributeValueTests.swift
```diff
@@ -338,48 +342,150 @@ extension AttributeValueTests {
...
         let attr = AttributeValue.string("string")
         var tmpError: Error?
         do {
-            _ = try attr.isSubstring(of: ["invalidType"])
+            _ = try attr.isExactMatch(with: invalidValue, condition: conditionString, name: attributeKey)
         } catch {
             tmpError = error
         }
-        XCTAssertTrue(tmpError != nil)
+        XCTAssertEqual("[Optimizely][Error] " + tmpError!.localizedDescription, OptimizelyError.evaluateAttributeInvalidType(conditionString, invalidValue, attributeKey).localizedDescription)
```
jaeopt, line 375: These tests validate expected error "type"s, but not testing if the log messages are actually printed out.
We throw errors. At some point, SDK is expected to catch it and prints out the log message. 
We need validate that the expected error messages are actually logged.

yasirfolio3: Hi jae, I have added new tests in decision service to verify if thrown errors are logged properly using `MockLogger`. i have also kept the newly added tests for `UserAttribute` and `AttributeValue` to test error handling but have removed assertion of log messages from them.

Tests/OptimizelyTests-DataModel/AttributeValueTests.swift
```diff
@@ -346,11 +346,11 @@ extension AttributeValueTests {
         let attr = AttributeValue.others
         var tmpError: Error?
         do {
-            _ = try attr.isExactMatch(with: invalidValue, condition: conditionString, name: attributeKey)
+            _ = try attr.isExactMatch(with: invalidValue)
         } catch {
             tmpError = error
         }
-        XCTAssertEqual("[Optimizely][Error] " + tmpError!.localizedDescription, OptimizelyError.evaluateAttributeInvalidCondition(conditionString).localizedDescription)
+        XCTAssertNotNil(tmpError)
```
jaeopt: Can we change these non-nil checking to specific error type checking?

`
var tmpError: OptimizelyError?
...
        } catch {
            tmpError = error as? OptimizelyError
        }
... 
if case . evaluateAttributeInvalidCondition = tmpError {
    XCTAssert(true)
} else {
   XCTAssert(false)
}
`

Tests/OptimizelyTests-Common/DecisionServiceTests_Experiments.swift
```diff
@@ -116,8 +211,17 @@ class DecisionServiceTests_Experiments: XCTestCase {
     override func setUp() {
         super.setUp()
         
+        HandlerRegistryService.shared.binders.property?.removeAll()
+        let binder: Binder = Binder<OPTLogger>(service: OPTLogger.self).to { () -> OPTLogger? in
+            return self.mockLogger
+        }
+        HandlerRegistryService.shared.registerBinding(binder: binder)
```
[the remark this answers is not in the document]
msohailhussain: Jae, DataStore binder sets its own logger which is causing problem here. All logs go there that's why reset it here.

Tests/OptimizelyTests-DataModel/AttributeValueTests.swift
yasirfolio3: I removed specific error type checking here since we are doing the same in the new decisionService tests.

jaeopt: I understand they're redundant. but it's better to have type checking here. Without them, these tests look incomplete.
Recovering your old XCTAssertEqual will also work.

PR #346, opened 2020-07-16
feat(decide): add UserContext

jaeopt:
- add OptimizelyUserContext
- add setUserContext API


Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,72 @@
+//
```
yasirfolio3, line 1: nit: please remove.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,72 @@
...
+/// A struct for user contexts that the SDK will use to make decisions for.
+public struct OptimizelyUserContext {
+    weak var optimizely: OptimizelyClient?
+    var userId: String
+    var attributes: [String: Any]
+    
+    /// OptimizelyUserContext init
+    ///
+    /// - Parameters:
+    ///   - optimizely: an instance of OptimizelyClient to be used for decisions.
```
yasirfolio3, line 30:
```suggestion
    ///   - optimizely: An instance of OptimizelyClient to be used for decisions.
```

Tests/OptimizelyTests-Common/OptimizelyUserContextTests.swift
```diff
@@ -0,0 +1,103 @@
+//
```
yasirfolio3, line 1: nit: please remove.

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
```diff
@@ -0,0 +1,36 @@
...
+/// Options controlling flag decisions.
+public enum OptimizelyDecideOption {
+    /// disable decision event tracking.
+    case disableDecisionEvent
+    
+    /// returns decisions only for flags which are enabled.
+    case enabledOnly
+    
+    /// skip user profile service for decision.
+    case ignoreUPS
```
yasirfolio3, line 29: As per the TDD:
```suggestion
    case ignoreUserProfileService
```

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import Foundation
+
+/// Options controlling flag decisions.
+public enum OptimizelyDecideOption {
+    /// disable decision event tracking.
+    case disableDecisionEvent
+    
+    /// returns decisions only for flags which are enabled.
+    case enabledOnly
```
yasirfolio3, line 26: As per the TDD:
```suggestion
    case enabledFlagsOnly
```

Sources/Optimizely+Decide/OptimizelyDecision.swift
```diff
@@ -0,0 +1,56 @@
...
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+import Foundation
+
+/// A decision struct that the SDK makes for a flag key and a user context.
+public struct OptimizelyDecision {
+    
+    /// A variation key of the decision. This value will be nil when decision making fails.
```
yasirfolio3, line 22: 'A' in this case doesn't look right, same comment for all properties in this struct.
```suggestion
    /// Variation key of the decision. This value will be nil when decision making fails.
```

Sources/Optimizely+Decide/OptimizelyClient+Decide.swift
```diff
@@ -0,0 +1,41 @@
+/****************************************************************************
```
yasirfolio3, line 1:
```suggestion
/***************************************************************************
```
Same for all modified or new files.

yasirfolio3 (commented):
Please update header year's on remaining modified files.

Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -34,6 +34,18 @@ open class OptimizelyClient: NSObject {
             atomicConfig.property = newValue
         }
     }
+    
+    private var atomicUserContext = AtomicProperty<OptimizelyUserContext>()
```
yasirfolio3, line 38: is this still required?

Sources/Optimizely+Decide/OptimizelyDecision.swift
```diff
@@ -0,0 +1,56 @@
...
+    public let variationKey: String?
+    
+    /// The boolean value indicating if the flag is enabled or not.
+    public let enabled: Bool
+    
+    /// The collection of variables assocaited with the decision.
+    public let variables: OptimizelyJSON?
+    
+    /// The rule key of the decision.
+    public let ruleKey: String?
```
msohailhussain, line 32: should be String or int?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// - Parameters:
+    ///   - optimizely: An instance of OptimizelyClient to be used for decisions.
+    ///   - userId: The user ID to be used for bucketing.
+    ///   - attributes: A map of attribute names to current user attribute values.
+    public init(optimizely: OptimizelyClient,
+                userId: String,
+                attributes: [String: Any]? = nil) {
+        self.optimizely = optimizely
+        self.userId = userId
+        self.attributes = attributes ?? [:]
```
msohailhussain, line 37: Shouldn't we have a shallow copy?

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
```diff
@@ -0,0 +1,36 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+
+import Foundation
+
+/// Options controlling flag decisions.
+public enum OptimizelyDecideOption {
```
msohailhussain, line 21: I think this should conform to `RawOptionSet` protocol? What do you say?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    public func trackEvent(eventKey: String, eventTags:  [String: Any]? = nil) {
+    }
+}
+
+extension OptimizelyUserContext: Equatable {
+    
+    public static func ==(lhs: OptimizelyUserContext, rhs: OptimizelyUserContext) -> Bool {
+        return lhs.userId == rhs.userId &&
+            (lhs.attributes as NSDictionary).isEqual(to: rhs.attributes)
```
msohailhussain, line 68: Just curious `equatable` is valid for Any?

msohailhussain (changes requested):
please address

Sources/Optimizely/OptimizelyClient.swift
jaeopt: can be dropped (type inferred)

Sources/Optimizely+Decide/OptimizelyDecision.swift
jaeopt: String just like all other keys

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: Swift always assigned a copy for dictionary, so no need for extra copying.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: Equatable is not for Any. I think we can simplify this Equatable function here 👍

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
jaeopt: A good point. I'll take a look if it's a better choice.

Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -34,6 +34,18 @@ open class OptimizelyClient: NSObject {
...
+    var userContext: OptimizelyUserContext? {
```
thomaszurkan-optimizely, line 39: I'm confused.  There is a standalone user context and a user context in the optimizely client?  Where is this ever set?

thomaszurkan-optimizely (commented):
I had some general questions.

Sources/Optimizely/OptimizelyClient.swift
thomaszurkan-optimizely: As a suggestion, can you serialize the user context to a dictionary and store that here so that you avoid holding a user context that holds a client (even weak)? It might be useful it would now just serialize client.

 Or as a less sophisticated approach you could just hold default user id and attributes and keep creating user contexts.

Tests/OptimizelyTests-Common/DecisionListenerTests.swift
```diff
@@ -856,9 +856,10 @@ class FakeManager: OptimizelyClient {
                   logger: OPTLogger? = nil,
                   eventDispatcher: OPTEventDispatcher? = nil,
                   userProfileService: OPTUserProfileService? = nil,
-                  defaultLogLevel: OptimizelyLogLevel? = nil) {
+                  defaultLogLevel: OptimizelyLogLevel? = nil,
+                  defaultDecideOptions: [OptimizelyDecideOption]? = nil) {
         
-        super.init(sdkKey: sdkKey, logger: logger, eventDispatcher: eventDispatcher, userProfileService: userProfileService, defaultLogLevel: defaultLogLevel)
+        super.init(sdkKey: sdkKey, logger: logger, eventDispatcher: eventDispatcher, userProfileService: userProfileService, defaultLogLevel: defaultLogLevel, defaultDecideOptions: defaultDecideOptions)
```
thomaszurkan-optimizely, line 862: Couldn't this be at the user context level _defaultDecideOptions_?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,71 @@
...
+* See the License for the specific language governing permissions and      *
+* limitations under the License.                                           *
+***************************************************************************/
+    
+
+import Foundation
+
+/// A struct for user contexts that the SDK will use to make decisions for.
+public struct OptimizelyUserContext {
+    weak var optimizely: OptimizelyClient?
```
thomaszurkan-optimizely, line 22: since this is weak and could go out of scope, we can add a guard error now for optimizely is nil

thomaszurkan-optimizely (changes requested):
Let's add the error we talked about and maybe a couple of unit tests for it if possible.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: It works ok since NSDictionary.isEqual calls each component.isEqual. Tests are added to confirm.

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
jaeopt: Will keep it as an array. All of the extended feature of OptionSet is not used here. Array of a few options will keep it simple.

Sources/Optimizely+Decide/OptimizelyDecideOption.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// disable decision event tracking.
+    case disableDecisionEvent
+    
+    /// return decisions only for flags which are enabled.
+    case enabledFlagsOnly
+    
+    /// skip user profile service for decision.
+    case ignoreUserProfileService
```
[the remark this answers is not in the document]
jaeopt: It can be useful and considered to add later.

Sources/Optimizely/OptimizelyClient.swift
jaeopt: Keeping a user-context was not an intention. It was a left-over while we change the architecture to createUserContext. It's cleaned up.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: A good point. Will add the guard and tests in other PRs (#347, #348).

Tests/OptimizelyTests-APIs/OptimizelyClientTests_Decide.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    override func setUp() {
+        super.setUp()
+        
+        let datafile = OTUtils.loadJSONDatafile("api_datafile")!
+        optimizely = OptimizelyClient(sdkKey: OTUtils.randomSdkKey)
+        try! optimizely.start(datafile: datafile)
+    }
+   
+    func testCreateUserContext() {
```
msohailhussain, line 31: Please add one more test, in which after creating user context, we should modify attributes and see those changes should not be reflected. 

e.g.
attributes["country"] = "uk"
and check user.attributes[country] shouldn't reflect that change.

Tests/OptimizelyTests-Common/OptimizelyUserContextTests.swift
```diff
@@ -0,0 +1,157 @@
...
+    
+    func testOptimizelyUserContext_nilAttributes() {
+        let user = OptimizelyUserContext(optimizely: expOptimizely, userId: expUserId, attributes: nil)
+        
+        XCTAssert(user.optimizely == expOptimizely)
+        XCTAssert(user.userId == expUserId)
+        XCTAssert(user.attributes.count == 0)
+    }
+
+    func testOptimizelyUserContext_attributes() {
```
msohailhussain, line 56: Please add one more test, when you modify existing key/value already in the attributes provided at the creation time.

msohailhussain (approved):
lgtm once tests are added, i asked for.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -0,0 +1,77 @@
...
+    var userId: String
+    
+    var atomicAttributes: AtomicProperty<[String: Any]>
```
thomaszurkan-optimizely, line 25: I love that we are using AtomicProperty.  Do you think there is a chance of a lot of these being created?  The reason I ask is that the more atomic properties we create, the more potential thread resources are being used (i.e. each concurrent dispatch queue created takes up resources).  So, I would suggest that we change the AtomicProperty to use a global concurrent queue (which will scale threads as needed) to save resources.  I'm going to create a separate PR for that right now.

thomaszurkan-optimizely (approved):
LGTM

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: It's not expected that we have many user-context (one in most cases), so no need here. But it can be helpful for other cases.

PR #348, opened 2020-07-16
feat(decide): add decideAll API

jaeopt:
- add decideAll() API


Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -127,12 +127,53 @@ public class OptimizelyUserContext {
...
+    /// - Returns: A dictionary of all decision results, mapped by flag keys.
+    public func decideAll(keys: [String],
+                          options: [OptimizelyDecideOption]? = nil) -> [String: OptimizelyDecision] {
+        
+        guard let optimizely = self.optimizely, let _ = optimizely.config else {
+            logger.e(OptimizelyError.sdkNotReady)
+            return [:]
+        }
+        
+        guard keys.count > 0 else { return [:] }
```
msohailhussain, line 147: Just curious, when `keys` is nil, will it return `empty`?

msohailhussain (changes requested):
will review later. just have one question.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
thomaszurkan-optimizely: keys can't be nil (not optional)

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -131,12 +131,53 @@ public class OptimizelyUserContext {
...
-
+    
+    /// Returns a key-map of decision results for all active flag keys.
+    ///
+    /// - Parameters:
+    ///   - options: An array of options for decision-making.
+    /// - Returns: A dictionary of all decision results, mapped by flag keys.
     public func decideAll(options: [OptimizelyDecideOption]? = nil) -> [String: OptimizelyDecision] {
-        return [:]
+        guard let optimizely = self.optimizely, let config = optimizely.config else {
```
thomaszurkan-optimizely, line 173: I think we talked about adding the guard statement in the first pr since it can be true at that point.  But, this captures the question in the first PR (error on optimizely going nil).  Do we have a unit test for this test case?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -131,12 +131,53 @@ public class OptimizelyUserContext {
...
+            logger.e(OptimizelyError.sdkNotReady)
+            return [:]
+        }
+
+        let keys = config.getFeatureFlags().map{ $0.key }
```
thomaszurkan-optimizely, line 178: could this be done lazy or maybe cached so it is only done once per config (i.e.  maybe store config version.)?

Sources/Optimizely/OptimizelyJSON.swift
```diff
@@ -181,3 +181,9 @@ public class OptimizelyJSON: NSObject {
         return nil
     }
 }
+
+extension OptimizelyJSON {
+    public static func ==(lhs: OptimizelyJSON, rhs: OptimizelyJSON) -> Bool {
```
thomaszurkan-optimizely, line 186: Can we add some OptimizelyJSON equals unit tests with different values?

thomaszurkan-optimizely (changes requested):
Some nit suggestions.

thomaszurkan-optimizely (approved):
The object serializing and deserializing to json and map concern me from a performance standpoint.  But, other than that: LGTM

PR #349, opened 2020-07-20
feat(decide): add decide reasons

jaeopt:
- add decide reasons


thomaszurkan-optimizely (changes requested):
Is there a way that you can register the reason dictionary with the service handler and then just have the logger inject the reason dictionary?  or, just have the logger take the reason as an argument so that you don't need to have reasons added as another parameter on every method.

jaeopt:
It's an interesting alternative. I think the current implementation can be more robust for supporting concurrent decide calls. Also, even for registry method, we still need pass down the decision task identity for concurrency support.
I cleaned up reasons processing.

thomaszurkan-optimizely (commented):
This seems like it would be so much cleaner and easier if you extended the logger protocol to include reasons and then pass the reasons struct to the logger to append.

Tests/OptimizelyTests-Common/DecisionServiceTests_Experiments.swift
```diff
@@ -342,8 +342,8 @@ extension DecisionServiceTests_Experiments {
         // no attributes, but UserProfile data by (1) will return valid variatin
         
         variation = self.decisionService.getVariation(config: config,
-                                                      userId: kUserId,
                                                       experiment: experiment,
+                                                      userId: kUserId,
```
thomaszurkan-optimizely, line 346: why move the user id

thomaszurkan-optimizely (approved):
LGTM

Tests/OptimizelyTests-Common/DecisionServiceTests_Experiments.swift
jaeopt: It's just for consistency. The "getVariation" method has the params in different order from other similar methods.

PR #357, opened 2020-08-25
fix: add check to see that all dot notated areas are numbers.

thomaszurkan-optimizely:
## Summary
- Add a check to make sure that the dot notated prefix is a number.

### DIFF
--- a/Sources/Data Model/Audience/SemanticVersion.swift
+++ b/Sources/Data Model/Audience/SemanticVersion.swift
@@ -88,7 +88,7 @@ extension SemanticVersion {
             throw OptimizelyError.attributeFormatInvalid
         }
         var targetedVersionParts = targetPrefix.split(separator: ".")
-        guard targetedVersionParts.count == dotCount + 1 else {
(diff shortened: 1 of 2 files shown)


Sources/Data Model/Audience/SemanticVersion.swift
```diff
@@ -88,7 +88,7 @@ extension SemanticVersion {
             throw OptimizelyError.attributeFormatInvalid
         }
         var targetedVersionParts = targetPrefix.split(separator: ".")
-        guard targetedVersionParts.count == dotCount + 1 else {
+        guard targetedVersionParts.count == dotCount + 1 && targetedVersionParts.filter({$0.isNumber}).count == targetedVersionParts.count else {
```
msohailhussain, line 91: am just curious if we need condition written above. 
``` else if !versionParts[idx].isNumber {```

thomaszurkan-optimizely: during the evaluation after the suffix has been added it is correct to test if the version is not a number.  i thought about adding the if versionParts is not a number and target is a number.  But, it should really fail in the split as a invalid format.

PR #380, opened 2021-01-05
feat(decide): change decision reasons to be returned as a response tuple

jaeopt:


thomaszurkan-optimizely (approved):
Looks good.  Why don't we merge this into the other decide PR.  Also, as a note, don't use xcode formatter in big PRs.  Those should be separate.  Thanks.

PR #381, opened 2021-01-08
feat(decide): clone user-context before calling optimizely decide

jaeopt:
Clone the current user-context and pass it to optimizely.decide()
- user context can be changed while processing the decide request.
- with this fix, decide can return the copy of the original user-context as a part of the decision


msohailhussain (approved):
lgtm
please change copyright headers.

PR #386, opened 2021-01-30
fix(decide): change nil to NSNull for decision flag notification

jaeopt:
RuleKey or VariationKey can be nil in decision flag notification.
To avoid they are removed from decisionInfo map, change them to NSNull (consistent with other decision notifications).


msohailhussain (dismissed):
lgtm

msohailhussain (commented):
lgtm

msohailhussain (approved):

PR #387, opened 2021-02-12
fix: some crashes have come up because of logger injection.  the reason seems to be the the binder is being updated every inject

thomaszurkan-optimizely:
## Summary
There have been a small number of crashes reported and it appears to be the injection of loggers from multiple threads at the same time.  This is because the binder was updating the instance variable even when the binder was always suppling new instances (i.e. when not a singleton).

The fix is two fold:
1.  If it is not a singleton just return a new instance via the factory and return.
2. Convert Binders to structs so there is a copy on read and a copy on write.  This way, any changes to the binders are changes to a copy and then the updating of the binder goes through an atomic property so the setter is atomic.


Sources/Utils/HandlerRegistryService.swift
```diff
@@ -52,16 +52,29 @@ class HandlerRegistryService {
         
         let binderToUse = binders.property?[skLocal] ?? binders.property?[skGlobal]
         
+        func updateBinder(b : BinderProtocol) {
+            if binders.property?[skLocal] != nil {
+                binders.property?[skLocal] = b
+            } else {
+                binders.property?[skGlobal] = b
+            }
```
jaeopt, line 60: Don't we need synchronization to update binders dictionary?
Let's fix this unsafe dictionaries/arrays in other places as well.

jaeopt (changes requested):
It looks good for fixing the logger issue.
We still have unsafe dictionaries as other possible sync failure sources. Can we fix them all?

DemoSwiftApp/AppDelegate.swift
```diff
@@ -17,6 +17,18 @@
 import UIKit
 import Optimizely
 
+private final class CustomLogger: OPTLogger {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: No, I just wanted to make sure it wasn't a problem if it was private.

Sources/Utils/HandlerRegistryService.swift
thomaszurkan-optimizely: the binder set is an atomic property.  the binder is a copy and so this is an atomic operation.  the get and set are separate.  but, the again, since this is only an issue with a singleton that usually provides a instance with creation. it should not be an issue.  i don't want to slow everything down by constantly wrapping everything in a dispatch.

DemoSwiftApp/AppDelegate.swift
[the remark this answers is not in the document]
jaeopt: Then let's keep it internal and remove this extra copy (or the other one).

thomaszurkan-optimizely:
> It looks good for fixing the logger issue.
> We still have unsafe dictionaries as other possible sync failure sources. Can we fix them all?

Can you tell me where we are accessing unsafe dictionaries?  I would handle that with a synched dictionary not by augment ing the atomic property by the way.

Tests/OptimizelyTests-APIs/OptimizelyClientTests_Others.swift
```diff
@@ -399,5 +399,28 @@ class OptimizelyClientTests_Others: XCTestCase {
...
+                 for k in 0..<10 {
+                     let logger = OPTLoggerFactory.getLogger()
+                     logger.log(level: .info, message: "[LOGGER] [\(i)] \(k)")
+                 }
+                if i == 999 {
+                    exp.fulfill()
+                }
+            }
+            // need to let the main global queue run otherwise some remained queued.
+            Thread.sleep(forTimeInterval: 0.02)
```
jaeopt, line 421: This sleep delay breaks the test. 1000 async tasks not concurrent any more because this delay.
If we remove the sleep delay and using DispatchGroup for sync, this test fails because concurrency issue.

jaeopt (changes requested):
The test is broken and cannot catch the thread issue here.

Tests/OptimizelyTests-APIs/OptimizelyClientTests_Others.swift
thomaszurkan-optimizely: > This sleep delay breaks the test. 1000 async tasks not concurrent any more because this delay.
> If we remove the sleep delay and using DispatchGroup for sync, this test fails because concurrency issue.

Please add a test if you think it would break.  The test won't run all the queued in the global queue without the sleep.  So.....

jaeopt (approved):
Approved for logger bug fix. 
The other thread issue I raised can be separated and will be reviewed later instead of blocking this fix.

PR #395, opened 2021-03-12
(DO NOT REVIEW) validate PR 384

jaeopt:


Sources/Extensions/OptimizelyClient+Extension.swift
```diff
@@ -51,32 +51,34 @@ extension OptimizelyClient {
...
     ///         The recommended value is 10 * 60 secs (you can also set this to nil to use the recommended value).
     ///         Set this to 0 to disable periodic downloading.
     ///   - defaultLogLevel: default log level (optional. default = .info)
     ///   - defaultDecisionOptions: default decision optiopns (optional)
     public convenience init(sdkKey: String,
                             logger: OPTLogger? = nil,
                             eventDispatcher: OPTEventDispatcher? = nil,
+                            datafileHandler: OPTDatafileHandler? = nil,
                             userProfileService: OPTUserProfileService? = nil,
                             periodicDownloadInterval: Int?,
```
yasirfolio3, line 66: is this now required? since we are already allowing user to provide datafileHandler as a whole.

Sources/Customization/Protocols/OPTDatafileHandler.swift
```diff
@@ -25,18 +25,35 @@ public protocol OPTDatafileHandler {
...
+     */
+    func setPeriodicInterval(sdkKey: String, interval: Int)
+    
+    /**
+     Check if periodic polling has been set
+     
+     - Parameter sdkKey: sdk key of the datafile to download
+     - Returns: true if set
+     */
+    func hasPeriodicInterval(sdkKey: String) -> Bool
```
yasirfolio3, line 43: why is this needed? since this is only used internally by the sdk and i don't see it being used publicly for any use case.

Sources/Customization/Protocols/OPTDatafileHandler.swift
```diff
@@ -25,18 +25,35 @@ public protocol OPTDatafileHandler {
...
     - Returns: a valid datafile or null
      */
     func downloadDatafile(sdkKey: String) -> Data?
     
     /**
      Asynchronous download data file.
      - Parameter sdkKey: application context for download
+     - Parameter returnCacheIfNoChange: include a cached datafile in result when datafile not changed.
      - Parameter resourceTimeoutInterval: timeout in seconds to wait for resource.
      - Parameter completionHhandler:  listener to call when datafile download complete
```
yasirfolio3, line 58:
```suggestion
     - Parameter completionHandler:  listener to call when datafile download is completed.
```

Tests/OptimizelyTests-Common/DatafileHandlerTests.swift
```diff
@@ -303,6 +303,22 @@ class DatafileHandlerTests: XCTestCase {
         // finally remove the datafile when complete.
         try? FileManager.default.removeItem(at: fileUrl!)
     }
+    
+    func testSetPeriodicInterval() {
+        let handler = DefaultDatafileHandler()
+        XCTAssert(handler.timers.property!.keys.isEmpty)
+        
+        handler.setPeriodicInterval(sdkKey: "abc", interval: 123)
```
yasirfolio3, line 311: it would be great if we could test it for multiple sdkKeys.

yasirfolio3 (commented):
Overall the changes look good, have made some minor comments. Also, there seems to be an issue with unit tests, they're failing with `error: Segmentation fault: 11`.

Sources/Customization/Protocols/OPTDatafileHandler.swift
jaeopt: This is used by OptimizelyClient to determine if config will be updated on datafile download. Since we now open DatafileHandler, this interface is added.

Tests/OptimizelyTests-Common/DatafileHandlerTests.swift
jaeopt: @yasirfolio3 Agreed. We'll add them when it's fixed to fully support multiple sdkKeys.

Sources/Extensions/OptimizelyClient+Extension.swift
jaeopt: @yasirfolio3 It should be much easier to turn on background polling in this way instead of extending the full datafileHandler.

PR #403, opened 2021-05-03
chore: turn on BUILD_LIBRARY_FOR_DISTRIBUTION for all platforms

jaeopt:
BUILD_LIBRARY_FOR_DISTRIBUTION is turned on all platforms to support ABI when clients build own a binary and share within the team.
It's supposed to be fixed in #325, but it turns out not be set for all platforms.


msohailhussain (approved):
lgtm

PR #406, opened 2021-05-11
fix(handler-registry-service): add support for multiple clients

jaeopt:
Fix handler-registry-service to support multiple clients (sdkKeys).

- full thread-safety for additional concurrency requirements
- no resource conflicts for multiple sdkKeys support


Tests/OptimizelyTests-MultiClients/LoggerTests_MultiClients.swift
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import XCTest
+
+class LoggerTests_MultiClients: XCTestCase {
+
+    override func setUpWithError() throws {
```
msohailhussain, line 21: Can we add `base` class and put these two `overrided` methods in there and extend classes from that `base` class. I see these two methods at many locations.

msohailhussain (approved):
lgtm. just have small suggestion.

Tests/OptimizelyTests-MultiClients/LoggerTests_MultiClients.swift
jaeopt: Most test cases will use own tasks in these methods. I'll clean up these empty methods.

PR #408, opened 2021-05-21
fix(notification): add support for multiple clients

jaeopt:
Fix NotificationCenter to support multiple clients (sdkKeys).

- full thread-safety for additional concurrency requirements
- no resource conflicts for multiple sdkKeys support

### DIFF
--- a/Tests/OptimizelyTests-MultiClients/NotificationCenterTests_MultiClients.swift
+++ b/Tests/OptimizelyTests-MultiClients/NotificationCenterTests_MultiClients.swift
@@ -1,0 +1,250 @@
+//
+// Copyright 2021, Optimizely, Inc. and contributors 
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");  
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at   
+// 
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


Tests/OptimizelyTests-MultiClients/NotificationCenterTests_MultiClients.swift
```diff
@@ -0,0 +1,250 @@
...
+        self.lock.sync {}
+
+        XCTAssertEqual(counter.activate, 0)
+        XCTAssertEqual(counter.track, numThreads * numEventsPerThread)
+        XCTAssertEqual(counter.decision, numThreads * numEventsPerThread)
+        XCTAssertEqual(counter.datafileChange, numThreads * numEventsPerThread)
+        XCTAssertEqual(counter.logEvent, numThreads * numEventsPerThread)
+    }
+    
+    func testConcurrentAddRemove() {
```
yasirfolio3, line 87: can we update this test case to remove only a part of the notification listeners. For example if we have 10 log event listeners, just remove 1 or 2 of them? because the current implementation seems very similar to `testConcurrentAddClear`. This is help us assert the remaining unremoved listeners of the similar kind.

yasirfolio3 (approved):
LGTM, just 1 suggestion.

msohailhussain (approved):
lgtm

PR #420, opened 2021-07-13
fix(OptimizelyConfig): add more tests for experiment audiences serialization

jaeopt:
- add more tests for audience serialization (empty audiences, NOT conditions)

### DIFF
--- a/Tests/OptimizelyTests-DataModel/ExperimentTests.swift
+++ b/Tests/OptimizelyTests-DataModel/ExperimentTests.swift
@@ -170,6 +170,7 @@ extension ExperimentTests {
         ]
         
         let audiencesInput: [Any] = [
+            [],
             ["or", "1", "2"],
             ["and", "1", "2", "3"],
             ["not", "1"],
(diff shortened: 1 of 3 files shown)


Tests/OptimizelyTests-DataModel/ExperimentTests.swift
```diff
@@ -192,6 +196,8 @@ extension ExperimentTests {
             "\"us\" OR \"female\"",
             "(\"us\" OR \"female\") AND \"adult\"",
             "(\"us\" OR (\"female\" AND \"adult\")) AND (\"fr\" AND (\"male\" OR \"kid\"))",
+            "NOT (\"us\" AND \"female\")",
```
keppel2, line 199: Use `#"raw "string" which is in quotes"#`?

jaeopt: I didn't know about this trick :) I'll skip it in this case.

keppel2: Swift 5+, no big deal!

The-inside-man (approved):
lgtm

PR #423, opened 2021-08-03
fix(OptimizelyConfig): add more opt-config tests with duplicate keys

jaeopt:
Add more tests for OptimizelyConfig with duplicate experiment- and delivery-rule keys


msohailhussain (approved):
lgtm

PR #431, opened 2021-08-12
feat(ForcedDecisions): add forced-decisions APIs to OptimizelyUserContext

jaeopt:
## Summary
Add a set of new APIs for forced-decisions to OptimizelyUserContext:
- setForcedDecision
- getForcedDecision
- removeForcedDecision

## Test plan
- unit tests for the new APIs
- FSC tests with new test cases


Sources/Data Model/ProjectConfig.swift
```diff
@@ -122,6 +124,41 @@ class ProjectConfig {
...
+        }()
+        
+        self.flagVariationsMap = {
+            var map = [String: [Variation]]()
+            
+            flagRulesMap.forEach { (flagKey, rules) in
+                var variations = [Variation]()
+                rules.forEach { rule in
+                    rule.variations.forEach { variation in
+                        if !variations.map({ $0.id }).contains(variation.id) {
```
msohailhussain, line 152: why this condition is needed? don't we have unique variations inside rollouts and experiments?

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -295,71 +288,126 @@ class DefaultDecisionService: OPTDecisionService {
...
-        // Evaluate fall back rule / last rule now
-        let experiment = rolloutRules[rolloutRules.count - 1]
+        return DecisionResponse(result: nil, reasons: reasons)
+    }
+    
+    func getVariationFromExperimentRule(config: ProjectConfig,
+                                        flagKey: String,
+                                        rule: Experiment,
+                                        user: OptimizelyUserContext,
+                                        options: [OptimizelyDecideOption]? = nil) -> DecisionResponse<Variation> {
```
msohailhussain, line 319: what about adding `IgnoreForcedVariation` in `OptimizelyDecideOption` ?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -122,15 +135,105 @@ public class OptimizelyUserContext {
...
+        
+        if let index = savedDecisions.firstIndex(where: { $0.flagKey == flagKey && $0.ruleKey == ruleKey }) {
+            _ = savedDecisions.remove(at: index)
+            return true
+        }
+        
+        return false
+    }
+    
+    public func removeForcedDecisions() -> Bool {
```
msohailhussain, line 192: what about `removeAllForcedDecisions`?

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -122,15 +135,105 @@ public class OptimizelyUserContext {
...
+        guard let optimizely = self.optimizely, optimizely.config != nil else {
+            logger.e(OptimizelyError.sdkNotReady)
+            return false
+        }
+        
+        savedDecisions.removeAll()
+        return true
+    }
+    
+    func findForcedDecision(flagKey: String, ruleKey: String? = nil) -> String? {
```
msohailhussain, line 202: this method should return reason.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -122,15 +135,105 @@ public class OptimizelyUserContext {
...
+        return nil
+    }
+    
+    func findValidatedForcedDecision(flagKey: String,
+                                     ruleKey: String?,
+                                     options: [OptimizelyDecideOption]? = nil) -> DecisionResponse<Variation> {
+        let reasons = DecisionReasons(options: options)
+        
+        if let variationKey = findForcedDecision(flagKey: flagKey, ruleKey: ruleKey) {
+            if let variation = optimizely?.getFlagVariationByKey(flagKey: flagKey, variationKey: variationKey) {
```
msohailhussain, line 218: in the `else` of this and above one should have `debug` reasons.

Sources/Data Model/ProjectConfig.swift
```diff
@@ -122,6 +124,41 @@ class ProjectConfig {
             project.rollouts.forEach { map[$0.id] = $0 }
             return map
         }()
+        
+        self.flagRulesMap = {
+            var map = [String: [Experiment]]()
+            
+            project.featureFlags.forEach { flag in
+                var experiments = flag.experimentIds.compactMap { expId in
+                    return allExperiments.filter { $0.id == expId }.first
```
msohailhussain, line 133: do we have any `map` of experiments by ids, from where we can lookup. 
filtering every time for each experimentId may be a bit performance issue.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -122,15 +135,105 @@ public class OptimizelyUserContext {
...
     }
+    
+}
 
+// MARK: - ForcedDecisions
+
+extension OptimizelyUserContext {
+    
+    public func setForcedDecision(flagKey: String,
+                                  ruleKey: String? = nil,
```
msohailhussain, line 152: `ruleKey` can we add comments, that it is for both experimentRule or deliveryRule.

msohailhussain (commented):
few questions for clarification.

Sources/Data Model/ProjectConfig.swift
jaeopt: We need "all variations for flag" to check if forced-decision is valid and to get the forced variation from its key. Since we do not have such "all variations for flag" in datafile, this code tried to build "variations" for a flag by walking through variations of all experiement/delivery rules. The condition is to avoid redundant listing.

jaeopt: Added comments

Sources/Implementation/DefaultDecisionService.swift
jaeopt: An interesting idea. I also see a usage case for this option, but not convinced enough to add as a decide option. We can discuss it.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: This is for a simple lookup. The reasons are included in "findValidatedForcedDecision" with the result of "findForcedDecision".

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: I refactored to add the reasons for this else. "Else" one for the above is intentionally dropped, since all decisions are with no forced decisions by default, polluting debug messages.

Tests/OptimizelyTests-Batch-iOS/EventDispatcherTests_Batch.swift
```diff
@@ -648,7 +648,7 @@ extension EventDispatcherTests_Batch {
 
         dispatchMultipleEvents([(kUrlA, batchEventA)])
         
-        wait(for: [eventDispatcher.exp!], timeout: 3)
+        wait(for: [eventDispatcher.exp!], timeout: 10)
```
The-inside-man, line 651: Where do we get these timeout values from? Are they environment dependent?

Tests/OptimizelyTests-Common/DatafileHandlerTests.swift
```diff
@@ -191,15 +191,15 @@ class DatafileHandlerTests: XCTestCase {
...
                 seconds = Int(abs(now.timeIntervalSinceNow))
+                expection.fulfill()
             }
         }
         
         wait(for: [expection], timeout: 20)
         
         XCTAssert(count == 10)
-        XCTAssert(seconds == 10)
+        XCTAssert(seconds > 5 && seconds < 20)
```
yasirfolio3, line 202: Is this change because of flakiness? if so, i think we should fix it in a way that no future updates are required to this.

Tests/OptimizelyTests-Common/OptimizelyUserContextTests_ForcedDecisions.swift
```diff
@@ -0,0 +1,392 @@
...
+        let featureKey = "feature_1"
+
+        // flag-to-decision
+        
+        var user = optimizely.createUserContext(userId: kUserId)
+        _ = user.setForcedDecision(flagKey: featureKey,
+                                   variationKey: "invalid")
+        var decision = user.decide(key: featureKey, options: [.includeReasons])
+        
+        // invalid forced-decision will be ignored and regular decsion will return
```
yasirfolio3, line 121:
```suggestion
        // invalid forced-decision will be ignored and regular decision will return
```

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -122,15 +135,123 @@ public class OptimizelyUserContext {
...
 
+extension OptimizelyUserContext {
+    
+    /// Sets the forced decision (variation key) for a given flag and an optional rule.
+    /// - Parameters:
+    ///   - flagKey: A flag key.
+    ///   - ruleKey: An experiment or delivery rule key (optional).
+    ///   - variationKey: A variation key.
+    /// - Returns: true if the forced decision has been set successfully.
+    public func setForcedDecision(flagKey: String,
```
yasirfolio3, line 157: what if empty flagKey is passed here?

yasirfolio3 (approved):
LGTM, just some minor suggestions.

Tests/OptimizelyTests-Batch-iOS/EventDispatcherTests_Batch.swift
jaeopt: No rule for setting this up - just make it large enough to avoid false-negative but small enough to catch failure fast. It is supposed to be completed in less than a sec, but I see tests intermittently fail in travis, which has been slowed down recently. It should work with the increased timeout.

Tests/OptimizelyTests-Common/DatafileHandlerTests.swift
jaeopt: A good point. I like to have large enough range to avoid flakiness, but still want to keep it narrow enough for meaningful testing :)

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -17,7 +17,7 @@
 import Foundation
 
 struct FeatureDecision {
-    let experiment: Experiment
+    var experiment: Experiment?
```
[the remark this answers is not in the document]
jaeopt: We add forcedDecision for {flag -> variation}, no specific experiment determination. So, it should be allowed to nil.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
jaeopt: It'll be considered as just like other invalid (no-matching) flagKeys. Instead of filtering the invalid keys here, we validate the forced-decision mapping when used for decision.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -119,18 +132,126 @@ public class OptimizelyUserContext {
...
+    ///   - flagKey: A flag key.
+    ///   - ruleKey: An experiment or delivery rule key (optional).
+    /// - Returns: A variation key or nil if forced decisions are not set for the parameters.
+    public func getForcedDecision(flagKey: String, ruleKey: String? = nil) -> String? {
+        guard optimizely?.config != nil else {
+            logger.e(OptimizelyError.sdkNotReady)
+            return nil
+        }
+        
+        return findForcedDecision(flagKey: flagKey, ruleKey: ruleKey)
```
The-inside-man, line 186: How come this is broken into two parts? It seems like the functionality of findForcedDecision is the serves the purpose of the get function here just with that logic taken out? Would it make sense to just bring that logic into the getForcedDecision and use the get function where we are calling the find function since the get calls the find?

jaeopt: The core function is separated out just to add SDK-ready-check only to the public getForcedDecision API.

The-inside-man: Makes sense, thank you for explaining that!

Tests/OptimizelyTests-Common/OptimizelyUserContextTests_ForcedDecisions.swift
```diff
@@ -0,0 +1,392 @@
...
+                                      userProfileService: OTUtils.createClearUserProfileService())
+        decisionService = (optimizely.decisionService as! DefaultDecisionService)
+        ups = decisionService.userProfileService
+        try! optimizely.start(datafile: datafile)
+    }
+   
+    func testSetForcedDecision_flagToDecision() {
+        let featureKey = "feature_1"
+        let user = optimizely.createUserContext(userId: kUserId)
+        _ = user.setForcedDecision(flagKey: featureKey,
```
yavorona, line 43: Can we also confirm the expected result of this call?

jaeopt: A good catch. I added a separate test to validate return status for all calls.

jaeopt: We have test cases for invalid variation keys - "testSetForcedDecision_invalid()".
Other cases with invalid flagKey and ruleKey will be considered as mismatches, so no explicit testing.

Sources/Data Model/ProjectConfig.swift
```diff
@@ -122,6 +124,44 @@ class ProjectConfig {
...
+        }()
+        
+        // all variations for each flag
+        // - datafile does not contain a separate entity for this.
+        // - we collect variations used in each rule (experiment rules and delivery rules)
+        
+        self.flagVariationsMap = {
+            var map = [String: [Variation]]()
+            
+            flagRulesMap.forEach { (flagKey, rules) in
```
The-inside-man, line 151: Can we include this in the initial iteration through the flags when we build the flag rules map? To avoid looping though the same experiments/rollouts twice?

jaeopt: A good point. Refactored.

Sources/Optimizely+Decide/OptimizelyUserContext.swift
```diff
@@ -119,18 +132,126 @@ public class OptimizelyUserContext {
...
+    ///   - flagKey: A flag key.
+    ///   - ruleKey: An experiment or delivery rule key (optional).
+    /// - Returns: true if the forced decision has been removed successfully.
+    public func removeForcedDecision(flagKey: String, ruleKey: String? = nil) -> Bool {
+        guard optimizely?.config != nil else {
+            logger.e(OptimizelyError.sdkNotReady)
+            return false
+        }
+        
+        if let index = forcedDecisions.firstIndex(where: { $0.flagKey == flagKey && $0.ruleKey == ruleKey }) {
```
The-inside-man, line 200: Thoughts on implementing using a map of maps for storing the forcedDecisions? Lookup time would be quicker and we can avoid looping through elements.

'Map<String, Map<String, ForcedDecision>> forcedDecisions = new.......'

then lookup just using the two keys: 'forcedDecisions.get(flagKey).get(ruleKey)'

msohailhussain (approved):
lgtm

PR #446, opened 2022-01-20
Fix invalidate url sessions

andredsp:
## Summary

According to Apple, [we should avoid creating instances of `URLSession`](https://developer.apple.com/videos/play/wwdc2017-709/?time=1977):

> But if you do allocate URLSessions dynamically, remember to clean up afterwards. Either finish tasks and invalidate or invalidate and cancel. But if you don't clean up, you'll leak memory.

## Detail

In the project I'm working on, we got this weird error:
```
Error Domain=NSPOSIXErrorDomain
Code=28 
“No space left on device” 
UserInfo={_kCFStreamErrorCodeKey=28, _kCFStreamErrorDomainKey=1}
```

Googling that lead us to discover that "space" here is probably not referring to RAM or disk space, but that as of iOS 14 this error shows up when you have too many URLSessions going at once:

* [Stack Overflow](https://stackoverflow.com/questions/67318867/error-domain-nsposixerrordomain-code-28-no-space-left-on-device-userinfo-kcf)
* [Apple dev forums thread](https://developer.apple.com/forums/thread/679107)

## Test plan
- Count instances of `MockUrlSession` that were created but not called `finishTasksAndInvalidate()` before it's destroyed.
- Assert that all `MockUrlSession` are invalidated on every `tearDown` that uses `MockUrlSession`

## Issues
- N/A

### DIFF
--- a/Sources/Customization/DefaultDatafileHandler.swift
+++ b/Sources/Customization/DefaultDatafileHandler.swift
@@ -69,6 +69,8 @@ open class DefaultDatafileHandler: OPTDatafileHandler {
             }
             
             let session = self.getSession(resourceTimeoutInterval: resourceTimeoutInterval)
+            // without this the URLSession will leak, see docs on URLSession and https://stackoverflow.com/questions/67318867
+            defer { session.finishTasksAndInvalidate() }
             
             guard let request = self.getRequest(sdkKey: sdkKey) else { return }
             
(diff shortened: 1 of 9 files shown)


jaeopt:
@andredsp Thanks for your contribution. We'll take a look at it.

jaeopt (approved):
LGTM

Issue #448, opened 2022-02-21
Crashes in ProjectConfig.swift


robinvan0418:
Hi there,

we observed some crashes in our app after implementing the OptimizelySDK. It seems to be very similar to this closed issue: https://github.com/optimizely/swift-sdk/issues/421

This crash occurred on iOS 14.x and 15.x devices.
```
Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x2b50 objc_release + 16
1  Optimizely                     0x19008 outlined consume of Project? + 4386328584
2  Optimizely                     0x98d94 ProjectConfig.deinit + 19 (ProjectConfig.swift:19)
3  Optimizely                     0x98e10 ProjectConfig.__deallocating_deinit + 19 (ProjectConfig.swift:19)
4  libswiftCore.dylib             0x388cb4 _swift_release_dealloc + 56
5  Optimizely                     0x9d60 AtomicProperty.__deallocating_deinit + 19 (AtomicProperty.swift:19)
6  libswiftCore.dylib             0x388cb4 _swift_release_dealloc + 56
7  Optimizely                     0x63e60 @objc OptimizelyClient.__ivar_destroyer + 4386635360 (OptimizelyClient.swift:4386635360)
8  libobjc.A.dylib                0x7d28 object_cxxDestructFromClass(objc_object*, objc_class*) + 116
9  libobjc.A.dylib                0x4b74 objc_destructInstance + 80
10 libobjc.A.dylib                0xe1c8 _objc_rootDealloc + 80
11 Optimizely                     0x727ec partial apply for closure #1 in OptimizelyClient.sendImpressionEvent(experiment:variation:userId:attributes:flagKey:ruleType:enabled:) + 4386695148 (<compiler-generated>:4386695148)
12 libswiftCore.dylib             0x388cb4 _swift_release_dealloc + 56
13 Optimizely                     0x2da80 partial apply for closure #2 in closure #1 in DefaultDatafileHandler.startPeriodicUpdates(sdkKey:updateInterval:datafileChangeNotification:) + 4386413184 (<compiler-generated>:4386413184)
14 libswiftCore.dylib             0x388cb4 _swift_release_dealloc + 56
15 libsystem_blocks.dylib         0x17e8 _Block_release + 200
16 libdispatch.dylib              0x4670 _dispatch_client_callout + 20
17 libdispatch.dylib              0x12b70 _dispatch_main_queue_callback_4CF + 944
[...]

PR #449, opened 2022-03-07
fix(ProjectConfig): remove sources of race conditions in ProjectConfig

jaeopt:
## Summary
This fixes a couple of possible sources of race conditions in ProjectConfig.
- A logger is created thread-safe in a lazy mode. This fixes a bug that the reference for the logger instance was not fully protected in synchronization.
- The static observer of {projectId, revision} changes in datafile is not necessary since the event-batcher checks them for 
batch boundaries. It's cleaned up in this fix since it becomes a source of tricky race conditions. 

## Test plan
- Fix unit tests covering loggers and event-batcher.

## Issues
- [Issue #448](https://github.com/optimizely/swift-sdk/issues/448)

### DIFF
--- a/Sources/Customization/DefaultDatafileHandler.swift
+++ b/Sources/Customization/DefaultDatafileHandler.swift
@@ -21,11 +21,11 @@ open class DefaultDatafileHandler: OPTDatafileHandler {
     public var endPointStringFormat = "https://cdn.optimizely.com/datafiles/%@.json"
     
     // thread-safe lazy logger load (after HandlerRegisterService ready)
-    private var loggerInstance: OPTLogger?
+    private let threadSafeLogger = ThreadSafeLogger()
     var logger: OPTLogger {
-        return OPTLoggerFactory.getLoggerThreadSafe(&loggerInstance)
+        return threadSafeLogger.logger
     }
-        
+
     // the timers for all sdk keys are atomic to allow for thread access.
     var timers = AtomicProperty(property: [String: (timer: Timer?, interval: Int)]())
     
--- a/Sources/Customization/DefaultDatafileHandler.swift
+++ b/Sources/Customization/DefaultDatafileHandler.swift
@@ -1,5 +1,5 @@
 //
-// Copyright 2019-2021, Optimizely, Inc. and contributors
+// Copyright 2019-2022, Optimizely, Inc. and contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/Sources/Customization/DefaultEventDispatcher.swift
+++ b/Sources/Customization/DefaultEventDispatcher.swift
@@ -38,9 +38,9 @@ open class DefaultEventDispatcher: BackgroundingCallbacks, OPTEventDispatcher {
     }
         
     // thread-safe lazy logger load (after HandlerRegisterService ready)
-    private var loggerInstance: OPTLogger?
+    private let threadSafeLogger = ThreadSafeLogger()
     var logger: OPTLogger {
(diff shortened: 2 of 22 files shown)


Sources/Customization/DefaultDatafileHandler.swift
```diff
@@ -21,11 +21,11 @@ open class DefaultDatafileHandler: OPTDatafileHandler {
     public var endPointStringFormat = "https://cdn.optimizely.com/datafiles/%@.json"
     
     // thread-safe lazy logger load (after HandlerRegisterService ready)
-    private var loggerInstance: OPTLogger?
+    private let threadSafeLogger = ThreadSafeLogger()
     var logger: OPTLogger {
-        return OPTLoggerFactory.getLoggerThreadSafe(&loggerInstance)
+        return threadSafeLogger.logger
     }
```
yasirfolio3, line 27: Although the code looks more readable this way, but how is this more thread safe? just curious! Instead of passing instance to a function, we have it inside a separate class now but the overall functionality remains the same.

Sources/Customization/DefaultEventDispatcher.swift
```diff
@@ -272,31 +266,3 @@ extension DefaultEventDispatcher {
...
-
-// MARK: - Notification Observers
-
-extension DefaultEventDispatcher {
-    
-    func addProjectChangeNotificationObservers() {
-        observers.append(
-            NotificationCenter.default.addObserver(forName: .didReceiveOptimizelyProjectIdChange, object: nil, queue: nil) { [weak self] _ in
-                self?.logger.d("Event flush triggered by datafile projectId change")
-                self?.flushEvents()
```
yasirfolio3, line 284: after removing this, the previous events will still be flushed before the new ones come in?

Tests/OptimizelyTests-Batch-iOS/EventDispatcherTests_Batch.swift
```diff
@@ -233,8 +230,8 @@ extension EventDispatcherTests_Batch {
     }
 
     func testBatchingEventsWhenProjectIdsNotEqual() {
-        // projectId change will flush all pending events, so this scenario should not happen frequently.
-        // but still possible when previous flushes failed.
+        // Batching set a boundary when an event has a different projectId from previous events.
+        // The remaining events will be sent as a separate batch (see "testFlushEventsWhenBatchFails()" for that).
```
yasirfolio3, line 234: Does this mean that when new events arrive, those will be sent before the previous events? is this acceptable?

yasirfolio3 (changes requested):

Sources/Customization/DefaultDatafileHandler.swift
jaeopt: "&loggerInstance" in old line 26 is read in outside of thread-safe sync (and passed to sync), so potentially can be a source of a race condition.

Sources/Customization/DefaultEventDispatcher.swift
jaeopt: The old events still remain in the batch queue. New events will append to the end of the queue. When flushed (timeout or full), batcher will group them to dispatch. If it sees a boundary of revision or projectId, it'll split into other groups. All dispatched in the correct order.

Tests/OptimizelyTests-Batch-iOS/EventDispatcherTests_Batch.swift
jaeopt: New events will be appended to the end of the batch queue. They'll be batched (with a projectId and revision boundaries) in FIFO order, so the dispatch order is preserved.

Issue #244, opened 2019-06-28
Revenue and Value for event tags is set as a AttributeValue which supports strings and bool.


thomaszurkan-optimizely:
these values can only be numbers.  they are set as attribute values which allow for strings and booleans.

jaeopt:
Fixed in https://github.com/optimizely/swift-sdk/pull/467

PR #467, opened 2022-09-28
fix: change the types of event revenue and value fields

jaeopt:
## Summary
The revenue and value fields of events are changed to integer and float types. 
- These changes are for internal only. There will be no change visible to users.
- They were implemented as AttributeValue type (the core types were validated when extracted).
 
## Test plan
- Pass existing unit and FSC tests

## Issues
- FSSDK-8562
- https://github.com/optimizely/swift-sdk/issues/244


msohailhussain (dismissed):
lgtm

msohailhussain (approved):
lgtm

PR #470, opened 2022-10-28
fix(ats): enhance or fix ODP integration

jaeopt:
## Summary
- When a vuid is used as a userId (UserContext is created without userId), identifyUser event should not include redundant vuids in event identifiers.

## Test plan
unit tests

## Issues
- FSSDK-8413


msohailhussain (approved):
lgtm

PR #477, opened 2023-02-02
fix(ats): remove odpConfig from OdpEventManager/OdpSegmentManager init params

jaeopt:
## Summary
OdpConfig is removed from OdpEventManager/OdpSegmentManager init params, which will be better for supporting custom OdpManager.

## Test plan
- Unit tests

## Issues
- [FSSDK-8843](https://jira.sso.episerver.net/browse/FSSDK-8843)


yasirfolio3 (approved):
LGTM

PR #512, opened 2023-06-16
[FSSDK-9433]: Adds support to override sdkName and sdkVersion for events

yasirfolio3:
## Summary
- Adds support to override sdkName and sdkVersion for events necessary for flutter.

## Test plan
- New unit tests added.
- All previous tests should pass

## Issues
- [FSSDK-9433](https://jira.sso.episerver.net/browse/FSSDK-9433)

### DIFF
--- a/Sources/Optimizely/OptimizelyClient.swift
+++ b/Sources/Optimizely/OptimizelyClient.swift
@@ -54,7 +54,7 @@ open class OptimizelyClient: NSObject {
     var logger: OPTLogger!
     var eventDispatcher: OPTEventDispatcher?
     public var datafileHandler: OPTDatafileHandler?
-    
+        
     // MARK: - Default Services
     
     var decisionService: OPTDecisionService!
@@ -76,7 +76,6 @@ open class OptimizelyClient: NSObject {
     ///   - defaultLogLevel: default log level (optional. default = .info)
     ///   - defaultDecisionOptions: default decision options (optional)
     ///   - settings: SDK configuration (optional)
-    ///   - clientName: clientName for impression and conversion events (optional)
     public init(sdkKey: String,
                 logger: OPTLogger? = nil,
                 eventDispatcher: OPTEventDispatcher? = nil,
@@ -90,7 +89,7 @@ open class OptimizelyClient: NSObject {
         self.sdkKey = sdkKey
         self.sdkSettings = settings ?? OptimizelySdkSettings()
         self.defaultDecideOptions = defaultDecideOptions ?? []
-        
+
         super.init()
         
         self.odpManager = odpManager ?? OdpManager(sdkKey: sdkKey,
@@ -209,7 +208,7 @@ open class OptimizelyClient: NSObject {
     func configSDK(datafile: Data) throws {
         do {
             self.config = try ProjectConfig(datafile: datafile)
-            
+                        
             datafileHandler?.startUpdates(sdkKey: self.sdkKey) { data in
                 // new datafile came in
                 self.updateConfigFromBackgroundFetch(data: data)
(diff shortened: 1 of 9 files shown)


Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -84,12 +86,13 @@ open class OptimizelyClient: NSObject {
                 odpManager: OdpManager? = nil,
                 defaultLogLevel: OptimizelyLogLevel? = nil,
                 defaultDecideOptions: [OptimizelyDecideOption]? = nil,
-                settings: OptimizelySdkSettings? = nil) {
+                settings: OptimizelySdkSettings? = nil,
+                clientName: String? = nil) {
```
jaeopt, line 90: Can we move this new option into `OptimizelySdkSettings`? It's not a common configuration usage.

jaeopt: We also need to config "sdkVersion" as well.
We can use "sdkName" and "sdkVersion" to be consistent with android-sdk?

Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -744,7 +750,7 @@ open class OptimizelyClient: NSObject {
             throw OptimizelyError.eventKeyInvalid(eventKey)
         }
         
-        sendConversionEvent(eventKey: eventKey, userId: userId, attributes: attributes, eventTags: eventTags)
+        sendConversionEvent(eventKey: eventKey, userId: userId, attributes: attributes, eventTags: eventTags, clientName: self.clientName)
```
jaeopt, line 753: I think we can change `Utils.sdkVersion` and `Utils.swiftSdkClientName` to update when customize. In that way, we do not change these apis.

Sources/Optimizely/OptimizelyClient.swift
```diff
@@ -285,7 +289,8 @@ open class OptimizelyClient: NSObject {
                             attributes: attributes,
                             flagKey: "",
                             ruleType: Constants.DecisionSource.experiment.rawValue,
-                            enabled: true)
+                            enabled: true,
+                            clientName: clientName)
```
jaeopt, line 293: Same as before. If we change `Utils.sdkVersion`, no need for this change.

jaeopt (changes requested):
Can we consider to change `Utils.sdkVersion` and `Utils.swiftSdkClientName` to reflect when customized? Then we may not need the changes in other codes.

Tests/OptimizelyTests-Common/BatchEventBuilderTests_Events.swift
```diff
@@ -28,6 +28,8 @@ class BatchEventBuilderTests_Events: XCTestCase {
     let datafile = OTUtils.loadJSONDatafile("api_datafile")!
     
     override func setUp() {
+        Utils.sdkVersion = OPTIMIZELYSDKVERSION
+        Utils.swiftSdkClientName = "swift-sdk"
```
jaeopt, line 32: Do we need this lines for testing? It may be more interesting without them to check if all works fine with default config.

jaeopt (changes requested):
Looks great! 
We also need a test case validating ODPEvent which includes sdk name and version.

Tests/OptimizelyTests-Common/BatchEventBuilderTests_Events.swift
yasirfolio3: Multiple tests started failing since these are global variables and the changes were reflected on all other unit tests. Some of our unit tests assert for swiftSdkClientName to be "swift-sdk".

jaeopt (approved):
LGTM

PR #518, opened 2023-08-17
[FSSDK-9600] Fix: Atomic property deadlocks issue 

muzahidul-opti:
## Summary
- Implemented RecurrsiveLock instead of DispatchQueue for AtomicProperty 

## Test plan
 - New test case added
 - The previous test should pass

## Issues
https://jira.sso.episerver.net/browse/FSSDK-9600


muzahidul-opti:
> Hello @muzahidul-opti - Any updates about this issue? We have forked the SDK and included this fix, and we haven't had any problems since then, but we don't feel very comfortable with that approach. I wonder if there's any other issue/concern that prevents this to be shipped in a new SDK release?

> Hi @lchinigioli thanks for your patience. We are investigating the issue. Will get back to you soon.

PR #541, opened 2024-01-12
[FSSDK-9945] fix: privacy manifest file value for required reason API

muzahidul-opti:
## Summary
- Current value was supplied by us, but the value should align with the list [CA92.1, 1C8F.1, C56D.1, AC6B.1] [Ref](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

## Test plan
- N/A

## Issues
- [FSSDK-9945](https://jira.sso.episerver.net/browse/FSSDK-9945)


jaeopt (approved):
LGTM

PR #543, opened 2024-01-17
Add PrivacyInfo inside resources in Swift Package

charlyliu:
## Summary
- As far as I know, PrivacyInfo needs to be as a resource in order to get it in the application. Same as in Cocoapods but for SPM.

## Issues
- Not opened issue

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,11 @@ let package = Package(
                  targets: ["Optimizely"])
     ],
     targets: [
-        .target(name: "Optimizely", path: "Sources")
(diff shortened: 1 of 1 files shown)


muzahidul-opti:
@charlyliu Thanks for your contribution! To incorporate your changes `swift-tools-version `also needs to be updated. Instead of going through the signing and testing process for the approval, we want to add this change into one of our PRs. If you do not mind, we can close the PR.

charlyliu:
Sure, I will close it. I also added a comment in your PR.

PR #564, opened 2024-11-21
[FSSDK-10761] fix: passed fsc for VUID optln

muzahidul-opti:
## Summary
- Passed new fsc test cases for VUID optln

## Test plan
- All test case should be passed

## Issues
- [FSSDK-10761](https://jira.sso.episerver.net/browse/FSSDK-10761)

### DIFF
--- a/Sources/ODP/OdpManager.swift
+++ b/Sources/ODP/OdpManager.swift
@@ -81,13 +81,6 @@ public class OdpManager {
                                                completionHandler: completionHandler)
     }
     
-    func sendInitializedEvent() {
-        guard let vuid = self.vuid else {
-            return
-        }
-        eventManager?.sendInitializedEvent(vuid: vuid)
-    }
-    
     func identifyUser(userId: String) {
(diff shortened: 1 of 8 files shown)


raju-opti (approved):
lgtm

Sources/ODP/OdpManager.swift
```diff
@@ -80,6 +81,12 @@ public class OdpManager {
                                                completionHandler: completionHandler)
     }
     
+    func sendInitializedEvent(vuid: String) throws {
+        guard enabled else { throw OptimizelyError.odpNotEnabled }
+        guard odpConfig.eventQueueingAllowed else { throw OptimizelyError.odpNotIntegrated }
```
jaeopt, line 86: Does this throw errors for default action (initailized) when ODP is not used? Can you double check that clients not using ODP not see any error messages from this?

jaeopt (commented):
Can you see my comments and confirm?

PR #596, opened 2025-06-04
Update DecisionInfo.swift

muzahidul-opti:
## Summary
- The "what"; a concise description of each logical change
- Another change

The "why", or other context.

## Issues
- "THING-1234" or "Fixes #123"

### DIFF
--- a/Sources/Implementation/DecisionInfo.swift
+++ b/Sources/Implementation/DecisionInfo.swift
@@ -140,17 +140,17 @@ struct DecisionInfo {
             decisionInfo[Constants.DecisionInfoKeys.variationKey] = variation?.key ?? NSNull()      // keep key in the map even with nil value
             decisionInfo[Constants.DecisionInfoKeys.ruleKey] = ruleKey ?? NSNull()                  //
             decisionInfo[Constants.DecisionInfoKeys.reasons] = reasons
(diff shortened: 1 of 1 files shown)


Sources/Implementation/DecisionInfo.swift
```diff
@@ -140,17 +140,17 @@ struct DecisionInfo {
...
             decisionInfo[Constants.DecisionInfoKeys.ruleKey] = ruleKey ?? NSNull()                  //
             decisionInfo[Constants.DecisionInfoKeys.reasons] = reasons
-        }
-        
-        decisionInfo[Constants.DecisionInfoKeys.decisionEventDispatched] = decisionEventDispatched
+            decisionInfo[Constants.DecisionInfoKeys.decisionEventDispatched] = decisionEventDispatched
         
-        if let expId = experiment?.id {
-            decisionInfo[Constants.ExperimentDecisionInfoKeys.experimentId] = expId
+            if let expId = experiment?.id {
```
FarhanAnjum-opti, line 145: Can we set the value here like we did for variationKey and ruleKey as above? This key should be in the map for nil value also.

muzahidul-opti: Adjust the changes

PR #602, opened 2025-07-01
[FSSDK-11172] feat: update decision service to handle CMAB

muzahidul-opti:
## Summary
- Update decision service to handle CMAB
- Add async support for decide apis
- Refractor default bucketer to handle CMAB traffic allocation

## Issues
- FSSDK-11172

### DIFF
--- a/Sources/Implementation/DefaultDecisionService.swift
+++ b/Sources/Implementation/DefaultDecisionService.swift
@@ -87,10 +87,18 @@ class DefaultDecisionService: OPTDecisionService {
                                               experiment: Experiment,
                                               user: OptimizelyUserContext,
                                               bucketingId: String,
+                                              opType: OPType,
                                               options: [OptimizelyDecideOption]?) -> DecisionResponse<VariationDecision> {
         let reasons = DecisionReasons(options: options)
         guard let cmab = experiment.cmab else {
             logger.e("The experiment isn't a CMAB experiment")
+            return DecisionResponse(result: nil, reasons: reasons)
+        }
+        
+        guard opType == .async else {
+            let info = LogMessage.cmabNotSupportedInSyncMode
+            logger.w(info)
+            reasons.addInfo(info)
             return DecisionResponse(result: nil, reasons: reasons)
         }
         
@@ -260,13 +268,11 @@ class DefaultDecisionService: OPTDecisionService {
             let bucketingId = getBucketingId(userId: userId, attributes: attributes)
             
             if experiment.isCmab {
-                if opType == .sync {
-                    /// fixme
-                }
                 let cmabDecisionResponse = getDecisionForCmabExperiment(config: config,
                                                                         experiment: experiment,
                                                                         user: user,
                                                                         bucketingId: bucketingId,
+                                                                        opType: opType,
                                                                         options: options)
                 reasons.merge(cmabDecisionResponse.reasons)
                 variationDecision = cmabDecisionResponse.result
--- a/Sources/Implementation/DefaultDecisionService.swift
+++ b/Sources/Implementation/DefaultDecisionService.swift
@@ -36,9 +36,6 @@ struct VariationDecision {
 
 typealias UserProfile = OPTUserProfileService.UPProfile
 
-let DEFAULT_CMAB_CACHE_TIMEOUT = 30 * 60 * 1000  // 30 minutes in milliseconds
-let DEFAULT_CMAB_CACHE_SIZE = 1000
-
 class DefaultDecisionService: OPTDecisionService {
     let bucketer: OPTBucketer
     let userProfileService: OPTUserProfileService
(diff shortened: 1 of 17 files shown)


Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -37,14 +58,87 @@ class DefaultDecisionService: OPTDecisionService {
...
+            logger.e("The experiment isn't a CMAB experiment")
+            return DecisionResponse(result: nil, reasons: reasons)
+        }
+        
+        guard opType == .async else {
+            let info = LogMessage.cmabNotSupportedInSyncMode
+            logger.w(info)
+            reasons.addInfo(info)
+            return DecisionResponse(result: nil, reasons: reasons)
+        }
```
jaeopt, line 103: Can we check if running thread, and reject if main thread?
CMAB can be called in sync decide inside of background thread?

jaeopt (changes requested):
It looks good. I have a suggestion to clean up .opType passing down. 
Let's discuss about the option.

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -37,14 +58,87 @@ class DefaultDecisionService: OPTDecisionService {
...
+        let bucketedResponse = (bucketer as? DefaultBucketer)?.bucketToEntityId(config: config, experiment: experiment, bucketingId: bucketingId, trafficAllocation: [cmabTrafficAllocation])
+        
+        if let _reasons = bucketedResponse?.reasons {
+            reasons.merge(_reasons)
+        }
+
+        let entityId = bucketedResponse?.result ?? ""
+        
+        // this means the user is not in the cmab experiment
+        if entityId != dummyEntityId {
```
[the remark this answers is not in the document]
muzahidul-opti: Done

Sources/Optimizely+Decide/OptimizelyClient+Decide.swift
```diff
@@ -166,6 +221,32 @@ extension OptimizelyClient {
...
+                        completion: @escaping DecideForKeysCompletion)  {
+        
+        decisionQueue.async {
+            guard let config = self.config else {
+                self.logger.e(OptimizelyError.sdkNotReady)
+                completion([:])
+                return
+            }
+            
+            let decision = self.decide(user: user, keys: config.featureFlagKeys, options: options,  opType: .async, ignoreDefaultOptions: false)
```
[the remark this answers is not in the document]
muzahidul-opti: I found some discrepancy here. `ignoreDefaultOptions` false for decideForKey but true for `decideForkeys`. But we can update to true.

Sources/Implementation/DefaultDecisionService.swift
muzahidul-opti: We can do that for safety, but CMAB can't be called inside sync decide.

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -456,12 +442,14 @@ class DefaultDecisionService: OPTDecisionService {
...
                                                                       user: user,
                                                                       userProfileTracker: userProfileTracker,
-                                                                      opType: opType,
+                                                                      isAsync: isAsync,
                                                                       options: options)
                 reasons.merge(decisionResponse.reasons)
                 if let result = decisionResponse.result {
-                    let featureDecision = FeatureDecision(experiment: experiment, variation: result.variation, source: Constants.DecisionSource.featureTest.rawValue, cmabUUID: result.cmabUUID)
-                    return DecisionResponse(result: featureDecision, reasons: reasons)
+                    if result.cmabError || result.variation != nil {
```
jaeopt, line 449: Is this correct?  should be `NOT result.cmabError`? Any case variation is not nil when cmabError is true?
```suggestion
                    if result.variation != nil {
```

Sources/Implementation/DefaultDecisionService.swift
```diff
@@ -80,24 +67,26 @@ class DefaultDecisionService: OPTDecisionService {
...
+                                              isAsync: Bool,
                                               options: [OptimizelyDecideOption]?) -> DecisionResponse<VariationDecision> {
         let reasons = DecisionReasons(options: options)
         guard let cmab = experiment.cmab else {
             logger.e("The experiment isn't a CMAB experiment")
             return DecisionResponse(result: nil, reasons: reasons)
         }
         
-        guard opType == .async else {
+        guard isAsync else {
```
jaeopt, line 87: Can we add comments why we do not take the alternative solution:
1. check the current thread.
2. reject if running in main thread.
This solution will work fine but we're concerned this may be confusing when users call flag decision with cmab
- when called from main thread, we'll skip cmab with error logging
- when called from background thread, we'll include cmab result
So the decision can be different.

jaeopt (changes requested):
Look good! A couple of small changes.

Sources/Implementation/DefaultDecisionService.swift
muzahidul-opti: > Is this correct? should be `NOT result.cmabError`? Any case variation is not nil when cmabError is true?
> 
> Suggested change
>                     if result.cmabError || result.variation != nil {
>                     if result.variation != nil {

This is intentional: when a CMAB error occurs, we return a decision with a nil variation.

muzahidul-opti: Add comment about cmab error handling

jaeopt (approved):
LGTM

PR #604, opened 2025-07-14
[FSSDK-11690] chore: remove layer id from holdout

muzahidul-opti:
## Summary
- Since the layer id is not required for HO, it has been removed.

## Test plan
- All existing test cases should be passed

## Issues
- FSSDK-11690

### DIFF
--- a/Sources/Data Model/Holdout.swift
+++ b/Sources/Data Model/Holdout.swift
@@ -66,7 +66,6 @@ extension Holdout: Equatable {
         lhs.key == rhs.key &&
         lhs.status == rhs.status &&
         lhs.variations == rhs.variations &&
-        lhs.layerId == rhs.layerId &&
         lhs.trafficAllocation == rhs.trafficAllocation &&
         lhs.audienceIds == rhs.audienceIds &&
         lhs.audienceConditions == rhs.audienceConditions &&


Sources/Data Model/Holdout.swift
```diff
@@ -65,8 +65,8 @@ extension Holdout: Equatable {
         return lhs.id == rhs.id &&
         lhs.key == rhs.key &&
         lhs.status == rhs.status &&
-        lhs.layerId == rhs.layerId &&
         lhs.variations == rhs.variations &&
+        lhs.layerId == rhs.layerId &&
```
jaeopt, line 69: Can't remove this line?

Sources/Data Model/Holdout.swift
```diff
@@ -36,20 +35,21 @@ struct Holdout: Codable, ExperimentCore {
...
     enum CodingKeys: String, CodingKey {
-        case id, key, status, layerId, variations, trafficAllocation, audienceIds, audienceConditions, includedFlags, excludedFlags
+        case id, key, status, variations, trafficAllocation, audienceIds, audienceConditions, includedFlags, excludedFlags
     }
     
     var variationsMap: [String: OptimizelyVariation] = [:]
     // replace with serialized string representation with audience names when ProjectConfig is ready
     var audiences: String = ""
+    // Not necessary for HO
+    var layerId: String = ""
```
jaeopt, line 45: Can't we remove this?

jaeopt (changes requested):

Sources/Data Model/Holdout.swift
muzahidul-opti: `ExperimentCore` protocol requires the implementation of `layerId`. There are two ways: we can either provide a default value via protocol extension or implement layerId explicitly in each confirming type. The second approach is much clearer to me.

PR #630, opened 2026-04-20
[FSSDK-12346] [Don't Merge] fix: upgrade NWPathMonitor to iOS 17.0 with fail-safe initialization

muzahidul-opti:
## Summary
- Update availability requirement from iOS 12.0 to iOS 17.0
- Replace force cast with safe optional guard
- Add defensive initialization to prevent null pointer crashes
- Network reachability gracefully disabled if monitor creation fails

This addresses rare crash issues on older iOS versions while providing fail-safe behavior for all supported platforms.

## Issues
- [FSSDK-12346](https://optimizely-ext.atlassian.net/browse/FSSDK-12346)

[FSSDK-12346]: https://optimizely-ext.atlassian.net/browse/FSSDK-12346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Utils/NetworkReachability.swift
+++ b/Sources/Utils/NetworkReachability.swift
@@ -52,28 +52,34 @@ class NetworkReachability {
     
     init(maxContiguousFails: Int? = nil) {
         self.maxContiguousFails = maxContiguousFails ?? NetworkReachability.defaultMaxContiguousFails
-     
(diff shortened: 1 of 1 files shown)


muzahidul-opti:
We have already released with this fix.

PR #631, opened 2026-04-30
[FSSDK-12394] chore:  local holdout logic behind FeatureGates flag

muzahidul-opti:
## Summary
- Added `FeatureGates.localHoldouts` static var (default `false`) to `Constants.swift`
- Wrapped both local holdout evaluation blocks in `DefaultDecisionService.swift` behind `if FeatureGates.localHoldouts`
- Created `BaseHoldoutTests` class that automatically enables the feature gate in setUp/tearDown
- Updated 6 holdout test classes to inherit from `BaseHoldoutTests`, removing duplicate gate setup code

## Test plan
- Ran all 6 holdout test suites (53 tests) — 0 failures
- Verified global holdout tests pass independently (not gated)
- Verified build succeeds via both `swift build` and `xcodebuild`
- Confirmed base test class properly enables/disables feature gate for all holdout tests

## Issues
- FSSDK-12394 (local holdouts feature gate)


raju-opti (dismissed):
Changes looks good. It would be good to add some tests verifying that feature gating is working properly, i.e holdouts is not evaluated when the flag is off.

muzahidul-opti:
> Changes looks good. It would be good to add some tests verifying that feature gating is working properly, i.e holdouts is not evaluated when the flag is off.

`FeatureGateTests_LocalHoldouts.swift` cases added

PR #513, opened 2023-06-20
feat: Swift concurrency, async-await support added

muzahidul-opti:
### Swift async-await support added for:
- Client initialization 
- ODP 

## Test plan
- Test cases added

## Issues
- [FSSDK-9419](https://jira.sso.episerver.net/browse/FSSDK-9419)

PR #636, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify event for single identifier

jaeopt:
## Summary

Fixes ODP identify event to only send when multiple valid identifiers exist. Previously, the SDK could send an identify event with just a single VUID when userId was nil, which provides no identity-linking value.

## Changes
- Added identifier count guard in `OdpEventManager.identifyUser` to skip event dispatch when fewer than 2 valid identifiers are present
- Added empty string check for both vuid and userId before adding to identifiers map
- Added debug-level logging when the identify event is skipped due to single identifier
- Updated test for nil userId case to expect no event queued (single identifier is now skipped)

## Jira Ticket
[FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

## Notes
- Reference implementation: Python SDK PR #513
- The count check is placed in OdpEventManager (where identifiers map is constructed) rather than OdpManager (which routes based on VUID detection)

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/ODP/OdpEventManager.swift
+++ b/Sources/ODP/OdpEventManager.swift
@@ -73,7 +73,7 @@ open class OdpEventManager {
         }
 
         guard identifiers.count >= 2 else {
-            logger.d("ODP identify event is not dispatched (only one identifier provided).")
+            logger.d("ODP identify event is not dispatched (fewer than 2 valid identifiers).")
             return
         }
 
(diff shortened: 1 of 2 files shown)


muzahidul-opti (commented):
I think, we need to figure out the right place to apply the condition.

muzahidul-opti (approved):
Block single indentifier events looks good to me.

PR #637, opened 2026-06-11
Revert local holdout feature gate logic

muzahidul-opti:
## Summary
- Reverted PR #631 (except CLAUDE.md documentation changes)
- Removed `FeatureGates.localHoldouts` flag from `Constants.swift`
- Unwrapped local holdout evaluation in `DefaultDecisionService.swift` so it runs unconditionally
- Removed `BaseHoldoutTests.swift` and `FeatureGateTests_LocalHoldouts.swift`
- Reverted holdout test classes back to inheriting from `XCTestCase`
- Removed corresponding `project.pbxproj` entries for deleted test files

Local holdout evaluation no longer needs to be gated behind a feature flag — it should run by default.

## Test plan
- Existing holdout tests continue to validate local holdout behavior
- No feature gate flag to toggle; holdout logic is always active

## Issues
- FSSDK-12394

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Sources/Utils/Constants.swift
```diff
@@ -16,10 +16,6 @@
 
 import Foundation
 
-struct FeatureGates {
```
raju-opti, line 19: can we keep this struct for future reference and just remove the field, maybe add a comment at the top about expected usage?

raju-opti (approved):
looks good. one comment

PR #642, opened 2026-06-24
[AI-FSSDK] [FSSDK-12813] normalize campaign_id, variation_id, and entity_id on decision events

jaeopt:
## Summary

Normalizes the three ID fields on outgoing **decision impression events** so the event-processing pipeline always receives pipeline-valid values:

- `decisions[].campaign_id`
- `decisions[].variation_id`
- `events[].entity_id`

Applies uniformly across decision types (experiment, feature test, rollout, holdout) — there is no per-type branching in the normalization path.

Per the updated spec ([FSSDK-12813](https://optimizely-ext.atlassian.net/browse/FSSDK-12813)):

- `campaign_id` and `entity_id` MUST be non-empty decimal-digit strings; when invalid (empty, whitespace, null, or non-numeric), the SDK falls back to the decision's `experiment_id`. Because both fields share the same source (`experiment.layerId`) and the same fallback, they MUST always hold the same normalized value on the wire (SC-006).
- `variation_id` MUST be a non-empty decimal-digit string OR explicit JSON `null`; when invalid (empty, whitespace, non-numeric), the SDK emits `null`.
- The fix lives in the event-builder layer only — no public API change, no datafile schema change, no new logs or warnings on the normalization path (FR-007).
- Events are never dropped as a result of normalization (FR-006).
- Conversion events derive `entity_id` from a different upstream source (the conversion event's own datafile `id`) and are out of scope (FR-010).

### Why generalize beyond holdouts?

The bug's loudest manifestation today is on holdout impression events (holdouts may legitimately lack `layerId`, so the invalid-`campaign_id` / `entity_id` path fires on every holdout event). Audited the SDK fleet and confirmed valid `campaign_id` / `entity_id` values are always non-empty numeric strings for non-holdout decisions in well-formed datafiles — so applying the same normalization uniformly is a true no-op for those paths today (verified by sibling regression suites, see below). The uniform rule simplifies the code and provides forward-compatibility for any future decision type or future datafile edge case.

## Changes

- `Sources/Implementation/Events/BatchEventBuilder.swift`
  - `createImpressionEvent` routes `campaign_id` / `variation_id` through `normalizeDecisionIds(...)` for every decision (no `isHoldout` branch).
  - `DispatchEvent.entityID` now receives the **normalized** `campaignId` instead of `rawCampaignId`, so `events[].entity_id` and `decisions[].campaign_id` share the same fallback (FR-009 / SC-006).
  - Adds `isValidNumericIdString(_:)` helper restricted to ASCII `[0-9]`.
- `Sources/Data Model/DispatchEvents/BatchEvent.swift` — `Decision.variationID` is `String?` with a custom `encode(to:)` that emits explicit JSON `null` (not `encodeIfPresent`) when nil, so the on-the-wire JSON carries `"variation_id": null` for the no-variation case.
- `Tests/OptimizelyTests-Common/BatchEventBuilderTests_HoldoutIds.swift` — 18 tests covering numeric-string validation, uniform `campaign_id` / `variation_id` normalization, FR-006 passthrough, JSON `null` serialization, and an end-to-end impression event assertion that `campaign_id` and `entity_id` both fall back to the holdout's `id` and are byte-equal. (Filename retained to keep `project.pbxproj` references stable; contents cover all decision types and all three ID fields.)
- `OptimizelySwiftSDK.xcodeproj/project.pbxproj` — registers the test file with `OptimizelyTests-Common-iOS` and `OptimizelyTests-Common-tvOS`.

## Verification

- New tests: **18/18 pass** (`BatchEventBuilderTests_HoldoutIds`)
- Regression suites: **40/40 pass** (`BatchEventBuilderTests_Events`, `BatchEventBuilderTests_Attributes`, `BatchEventBuilderTests_EventTags`, `BatchEventBuilderTests_Region`, `DecisionServiceTests_Holdouts`) — confirms the generalized normalization is a true no-op for existing valid non-holdout events (SC-003) and that existing impression-event assertions (`entity_id == campaign_id`) continue to hold.
- Run: `xcodebuild -workspace OptimizelySwiftSDK.xcworkspace -scheme OptimizelySwiftSDK-iOS -destination 'platform=iOS Simulator,name=iPhone 17' test`

## Jira Ticket

[FSSDK-12813](https://optimizely-ext.atlassian.net/browse/FSSDK-12813)

[FSSDK-12813]: https://optimizely-ext.atlassian.net/browse/FSSDK-12813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FSSDK-12813]: https://optimizely-ext.atlassian.net/browse/FSSDK-12813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/OptimizelySwiftSDK.xcodeproj/project.pbxproj
+++ b/OptimizelySwiftSDK.xcodeproj/project.pbxproj
@@ -2115,6 +2115,8 @@
 		98AC97F42DAE9685001405DD /* HoldoutConfigTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98AC97F22DAE9685001405DD /* HoldoutConfigTests.swift */; };
 		98AC98462DB7B762001405DD /* BucketTests_HoldoutToVariation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98AC98452DB7B762001405DD /* BucketTests_HoldoutToVariation.swift */; };
 		98AC98472DB7B762001405DD /* BucketTests_HoldoutToVariation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98AC98452DB7B762001405DD /* BucketTests_HoldoutToVariation.swift */; };
+		FF12813000000000FF000002 /* BatchEventBuilderTests_HoldoutIds.swift in Sources */ = {isa = PBXBuildFile; fileRef = FF12813000000000FF000001 /* BatchEventBuilderTests_HoldoutIds.swift */; };
+		FF12813000000000FF000003 /* BatchEventBuilderTests_HoldoutIds.swift in Sources */ = {isa = PBXBuildFile; fileRef = FF12813000000000FF000001 /* BatchEventBuilderTests_HoldoutIds.swift */; };
 		98AC98492DB8FC29001405DD /* DecisionServiceTests_Holdouts.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98AC98482DB8FC29001405DD /* DecisionServiceTests_Holdouts.swift */; };
 		98AC984B2DB8FFE0001405DD /* DecisionServiceTests_Holdouts.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98AC98482DB8FC29001405DD /* DecisionServiceTests_Holdouts.swift */; };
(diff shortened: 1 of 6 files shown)


Mat001 (dismissed):
lgtm

PR #646, opened 2026-06-26
[AI-FSSDK] [FSSDK-12834] Replace non-numeric holdout placeholder IDs in event-construction tests

jaeopt:
## Summary
Per FR-011 / SC-007 of the decision-event ID normalization spec, every pre-existing test that constructs an event payload must use valid decimal-digit string IDs. Non-numeric holdout placeholders implicitly relied on the pre-fix pass-through behavior and would either silently change their on-the-wire output after the normalization landed or assert the wrong thing.

## Changes
- Replace non-numeric holdout `id`, variation `id`, and trafficAllocation `entityId` placeholders with numeric-string IDs across all holdout-related test fixtures whose tests dispatch impression events (BatchEventBuilder, DecisionListener, decide-API, decide-reasons, local-holdout suites).
- Update the one event-level assertion that previously expected the non-numeric placeholder to fall through, so it now asserts the realistic numeric IDs and a numeric `variation_id` (matching the post-fix happy-path normalization).
- Leave the dedicated invalid-input-path tests in `BatchEventBuilderTests_HoldoutIds.swift` and the no-dispatch `DecisionServiceTests_Holdouts.swift` fixtures unchanged — the former intentionally exercises the fallback contract, the latter never reaches the event builder.

## Jira Ticket
[FSSDK-12834](https://optimizely-ext.atlassian.net/browse/FSSDK-12834)

[FSSDK-12834]: https://optimizely-ext.atlassian.net/browse/FSSDK-12834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Implementation/Events/BatchEventBuilder.swift
+++ b/Sources/Implementation/Events/BatchEventBuilder.swift
@@ -62,35 +62,23 @@ class BatchEventBuilder {
                                 dispatchEvents: [dispatchEvent])
     }
 
-    // MARK: - Decision Event ID Normalization (FSSDK-12813, FSSDK-12834)
+    // MARK: - Decision Event ID Normalization
 
-    /// Normalizes `campaign_id` and `variation_id` so every dispatched decision
-    /// event carries pipeline-valid values, regardless of decision type
-    /// (experiment, feature test, rollout, or holdout). See spec FR-001–FR-005.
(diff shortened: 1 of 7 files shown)


muzahidul-opti (approved):
LGTM

PR #647, opened 2026-06-27
[AI-FSSDK] [FSSDK-12839] Add no-op regression test for non-holdout decision id normalization

jaeopt:
## Summary
Locks in the byte-equivalent no-op behavior of decision-event id normalization for non-holdout decisions (experiments, feature tests, rollouts). The existing holdout tests cover the fallback path; this regression covers the happy path so a future refactor cannot silently start substituting valid IDs.

## Changes
- Add a unit-level regression test asserting that `normalizeDecisionIds` passes valid numeric `campaign_id` and `variation_id` through unchanged, and that `campaign_id` does not collapse to `experiment_id` when the raw value is already valid.

## Jira Ticket
[FSSDK-12839](https://optimizely-ext.atlassian.net/browse/FSSDK-12839)

[FSSDK-12839]: https://optimizely-ext.atlassian.net/browse/FSSDK-12839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/OptimizelyTests-Common/BatchEventBuilderTests_HoldoutIds.swift
+++ b/Tests/OptimizelyTests-Common/BatchEventBuilderTests_HoldoutIds.swift
@@ -275,4 +275,38 @@ class BatchEventBuilderTests_HoldoutIds: XCTestCase {
         XCTAssertEqual(campaignId, entityId,
                        "campaign_id and entity_id must hold the same normalized value")
     }
+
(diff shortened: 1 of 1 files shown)


jaeopt:
TESTING ONLY

PR #648, opened 2026-07-01
[AI-FSSDK] [FSSDK-12750] Use attribute id instead of key for CMAB prediction requests

raju-opti:
## Summary

The CMAB service was using attribute keys instead of attribute IDs when building the filtered attribute map for prediction requests. The CMAB prediction endpoint expects attribute IDs in the JSON payload, so this change corrects the map key from the attribute's key to its ID.

## Changes

- Updated the filtered attribute map in CmabService to use attribute ID as the dictionary key instead of attribute key
- Updated all related test assertions and cache hash computations to use attribute IDs
- Fixed a pre-existing bug in MockProjectConfig where an attribute was incorrectly added to attributeIdMap instead of attributeKeyMap

## Jira Ticket

[FSSDK-12750](https://optimizely-ext.atlassian.net/browse/FSSDK-12750)

[FSSDK-12750]: https://optimizely-ext.atlassian.net/browse/FSSDK-12750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/CMAB/CmabService.swift
+++ b/Sources/CMAB/CmabService.swift
@@ -190,7 +190,7 @@ class DefaultCmabService: CmabService {
         let cmabAttributeIds = cmab.attributeIds
         for attributeId in cmabAttributeIds {
             if let attribute = config.getAttribute(id: attributeId), let value = userAttributes[attribute.key] {
-                filteredUserAttributes[attribute.key] = value
+                filteredUserAttributes[attribute.id] = value
             }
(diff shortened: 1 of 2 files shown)


muzahidul-opti (approved):
LGTM

PR #650, opened 2026-07-08
[FSSDK-12346] fix: defer NWPathMonitor creation to avoid crash during multi-SDK startup

muzahidul-opti:
## Summary
- Deferred `NWPathMonitor` creation from `NetworkReachability.init` to first `shouldBlockNetworkAccess()` call
- Removed force casts (`as! NWPathMonitor`) in favor of local variable binding
- Added `monitorStarted` flag in `stop()` to prevent late monitor creation after teardown
- Added tests verifying monitor is not created during init and not created after `stop()`

In Flutter apps with multiple SDKs (Firebase, Datadog, MParticle) on iOS 16, `NWPathMonitor` was eagerly created during `DefaultEventDispatcher` stored-property initialization. This caused a null function pointer crash (`EXC_BAD_ACCESS` at `0x0`) due to dyld contention over the weak-linked Network framework during concurrent framework loading. The crash did not reproduce on iOS 17+ (improved dyld concurrency) or in sample apps (no competing SDKs).

By deferring `NWPathMonitor` creation to first use during event dispatch, the Network framework symbols are resolved well after app startup contention has settled.

## Test plan
- Added `testMonitorNotCreatedDuringInit` — confirms `monitor` is nil after init, non-nil after `shouldBlockNetworkAccess()`
- Added `testMonitorNotCreatedAfterStop` — confirms monitor stays nil if `stop()` was called before first use
- Updated `testReachabilityMonitoring` to explicitly trigger monitor start
- All existing `NetworkReachabilityTests` and `EventDispatcherRetryTests` pass

## Issues
- FSSDK-12346

### DIFF
--- a/Sources/Utils/NetworkReachability.swift
+++ b/Sources/Utils/NetworkReachability.swift
@@ -18,10 +18,11 @@ import Foundation
 import Network
 
 class NetworkReachability {
-    
+
     var monitor: AnyObject?
     let queue = DispatchQueue(label: "reachability")
-    
+    private var monitorStarted = false
+
     // the number of contiguous download failures (reachability)
     var numContiguousFails = 0
     // the maximum number of contiguous network connection failures allowed before reachability checking
@@ -33,7 +34,7 @@ class NetworkReachability {
     #else
     private var connected = true        // initially true for safety in production
     #endif
-    
+
     var isConnected: Bool {
         get {
             var result = false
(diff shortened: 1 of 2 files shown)


Sources/Utils/NetworkReachability.swift
```diff
@@ -49,45 +50,54 @@ class NetworkReachability {
...
-            (monitor as! NWPathMonitor).start(queue: queue)
+
+            pathMonitor.start(queue: queue)
         }
     }
-    
+
     func stop() {
+        monitorStarted = true
         if #available(macOS 10.14, iOS 12.0, watchOS 5.0, tvOS 12.0, *) {
```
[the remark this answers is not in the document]
muzahidul-opti: Suggestions added

Sources/Utils/NetworkReachability.swift
```diff
@@ -18,83 +18,102 @@ import Foundation
...
-    
-    func stop() {
-        if #available(macOS 10.14, iOS 12.0, watchOS 5.0, tvOS 12.0, *) {
-            guard let monitor = monitor as? NWPathMonitor else { return }
 
-            monitor.pathUpdateHandler = nil
-            monitor.cancel()
+    func stop() {
+        queue.sync {
+            monitorStarted = true
```
raju-opti, line 88: should this be set to true inside stop() method?

Sources/Utils/NetworkReachability.swift
```diff
@@ -18,83 +18,102 @@ import Foundation
...
+                let pathMonitor = NWPathMonitor()
+                self.monitor = pathMonitor
+                pathMonitor.pathUpdateHandler = { [weak self] path in
+                    self?._connected = (path.status == .satisfied)
+                }
+                monitorToStart = pathMonitor
             }
-            
-            (monitor as! NWPathMonitor).start(queue: queue)
+            monitorToStart?.start(queue: queue)
```
raju-opti, line 82: why not self.monitor.start()?

Sources/Utils/NetworkReachability.swift
muzahidul-opti: Yes — this is intentional. Setting `monitorStarted = true` in `stop()` prevents the monitor from being lazily re-created on the next `shouldBlockNetworkAccess()` call.

Without it, the sequence `stop()` → `shouldBlockNetworkAccess()` → `startMonitorIfNeeded()` would see `monitorStarted == false` and spin up a new `NWPathMonitor`, undoing the stop.

Note: `stop()` is only called from tests (never in production code) to disconnect live monitoring so tests can manually control `isConnected`. This flag ensures the monitor stays stopped for the rest of that test.

Sources/Utils/NetworkReachability.swift
muzahidul-opti: We intentionally avoid calling `self.monitor.start(queue: queue)` inside the `queue.sync` block to prevent a potential deadlock.

`NWPathMonitor.start(queue:)` schedules the `pathUpdateHandler` on the same serial `queue`. On some iOS versions, the handler can fire synchronously during `start()`. If that happens while we're inside `queue.sync` (already holding the queue), the handler dispatch would wait for the queue → deadlock.

By capturing the monitor into a local `monitorToStart` variable inside the lock, then calling `.start()` outside, we keep the critical section short (flag + creation + handler setup) and avoid any deadlock risk with the queue.

raju-opti (approved):

